PR #1, opened 2026-07-23
Build pinned Harbor runner image

LJXLJXLJX:
## 1. Why the change?

The Harbor control runner previously installed unpinned dependencies at workload startup, allowing runner versions to drift and making startup depend on package-index availability.

Harbor 0.20.0 may write an unquoted `[REDACTED]` value while serializing trial data, producing an invalid `result.json` ([upstream implementation](https://github.com/harbor-framework/harbor/blame/071281b3d931aafd6a5375fa7d5933e23054d784/src/harbor/trial/trial.py#L771)). Pin the runner to Harbor 0.18.0 and Opik 2.1.32 while preserving both Docker and host-direct launch modes.

## 2. Summary of the change

- Define `harbor==0.18.0` and `opik==2.1.32` in a shared runner requirements file.
- Install the pinned environment in the repository-built DinD image and create the same environment for host-direct runs during setup.
- Replace workload-time installation with fail-fast Python, package-version, and CLI validation.
- Fingerprint image build inputs so stale runner images are not silently reused.
- Use a Debian/glibc DinD base and binary wheels to avoid costly source builds.
- Keep the Harbor `--append-system-prompt` compatibility fix independent of tracing.
- Keep the nested Docker daemon on its Unix socket and allow container recreation without deleting the persistent Docker cache.

## 3. Other details

- Migrated onto the current `sii-system/agent-fleet` `main`.
- Harbor common Python tests, Claude compatibility tests, Harbor shell tests, DinD launcher tests, shell syntax checks, and `git diff --check` pass.
- A real SETA task completed the `/run_trial` path with a valid structured result, queue archival, and container cleanup.


LongHanyu (approved):
LGTM

PR #2, opened 2026-07-23
Start Harbor analyzer from run lifecycle

0407nancy:
## Why the change?

Harbor monitor already emits analyzer handovers, and the analyzer already supports follow mode, but they are not started under one run lifecycle. Users must launch the analyzer manually, which makes PID/log/reset behavior inconsistent with the monitor.

This PR migrates sii-system/sii-agent-fleet#95 to sii-system/agent-fleet.

## Summary of the change

- Add opt-in Harbor analyzer lifecycle settings in `env.sh`.
- Start the existing `analyzer_subagent.py --follow` after the monitor produces `analyzer-handover-latest.json`.
- Add run-scoped analyzer PID matching and reset cleanup.
- Drain analyzer handovers before shutdown in command and foreground zellij modes.
- Add a detached analyzer supervisor for detached zellij runs.
- Document the opt-in analyzer startup path.

## Other details

The analyzer remains disabled by default because it can consume model calls. The monitor still owns benchmark control decisions; the analyzer only consumes monitor handovers and writes reports.

Validation:
- `bash -n Agents/utils/common/Harbor/env.sh`
- `git diff --cached --check`

Not run:
- Harbor analyzer pytest tests because `pytest` is not installed in this environment.

### DIFF
--- a/Agents/utils/common/Harbor/start.sh
+++ b/Agents/utils/common/Harbor/start.sh
@@ -143,6 +143,172 @@ harbor_start_monitor_if_enabled() {
   ) 9>"$RUNTIME_DIR/harbor-monitor.lock"
 }
 
+harbor_start_analyzer_if_enabled() {
+  [[ "$HARBOR_ANALYZER_ENABLED" == "1" ]] || return 0
+  [[ "$ROLLOUT" != "1" ]] || return 0
+  if [[ "$HARBOR_MONITOR_ENABLED" != "1" ]]; then
+    echo "[ERROR] cannot start Harbor analyzer because Harbor monitor is disabled" >&2
+    return 1
+  fi
+
+  (
+  flock 9
+  if [[ -f "$HARBOR_ANALYZER_PID_FILE" ]]; then
+    local existing_pid
+    existing_pid="$(cat "$HARBOR_ANALYZER_PID_FILE" 2>/dev/null || true)"
+    if [[ -n "$existing_pid" ]] && kill -0 "$existing_pid" >/dev/null 2>&1; then
+      if harbor_analyzer_pid_matches_run "$existing_pid"; then
+        exit 0
+      fi
+      echo "[ERROR] refusing to replace unrelated process from $HARBOR_ANALYZER_PID_FILE: pid=$existing_pid" >&2
+      exit 1
+    fi
+    rm -f "$HARBOR_ANALYZER_PID_FILE"
+  fi
+
+  for _ in $(seq 1 50); do
+    [[ -f "$HARBOR_MONITOR_DIR/analyzer-handover-latest.json" ]] && break
+    sleep 0.1
+  done
+  if [[ ! -f "$HARBOR_MONITOR_DIR/analyzer-handover-latest.json" ]]; then
+    echo "[ERROR] cannot start Harbor analyzer before monitor handover exists" >&2
+    exit 1
+  fi
+
+  mkdir -p "$HARBOR_ANALYZER_OUTPUT_DIR" "$RUNTIME_DIR"
+  local analyzer_pid
+  case "$HARBOR_ANALYZER_MODE" in
+    handover-follow)
+      nohup setsid python3 "$SCRIPT_DIR/scripts/analyzer_subagent.py" \
+        --handover "$HARBOR_MONITOR_DIR/analyzer-handover-latest.json" \
+        --handoff-dir "$HARBOR_MONITOR_DIR/analyzer-handoffs" \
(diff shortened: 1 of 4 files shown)


Agents/utils/common/Harbor/start.sh
```diff
@@ -143,6 +143,172 @@ harbor_start_monitor_if_enabled() {
...
+}
+
+harbor_wait_for_analyzer_drain() {
+  [[ "$HARBOR_ANALYZER_ENABLED" == "1" && "$ROLLOUT" != "1" ]] || return 0
+  [[ -f "$HARBOR_ANALYZER_PID_FILE" ]] || return 0
+  local pid timeout deadline
+  pid="$(cat "$HARBOR_ANALYZER_PID_FILE" 2>/dev/null || true)"
+  [[ "$pid" =~ ^[0-9]+$ ]] || return 0
+  kill -0 "$pid" >/dev/null 2>&1 || return 0
+  timeout="${HARBOR_ANALYZER_DRAIN_TIMEOUT:-60}"
```
qiyuangong, line 281: [P1] Do not cap draining below configured analyzer work

`HARBOR_ANALYZER_TIMEOUT` defaults to 900 seconds and concurrency defaults to one, but this deadline stops pending handovers after 60 seconds. A valid final analysis can therefore be killed before its own timeout. Wait for the active analysis to finish, or derive a shutdown deadline that covers the configured work.

Agents/utils/common/Harbor/start.sh
```diff
@@ -143,6 +143,172 @@ harbor_start_monitor_if_enabled() {
...
+
+harbor_start_detached_analyzer_supervisor_if_enabled() {
+  [[ "$HARBOR_ANALYZER_ENABLED" == "1" && "$ROLLOUT" != "1" ]] || return 0
+  (
+    trap '' HUP
+    while harbor_monitor_is_running_for_run; do
+      sleep 5
+    done
+    harbor_wait_for_analyzer_drain
+    harbor_stop_analyzer || true
```
qiyuangong, line 308: [P1] Fence the detached supervisor to its analyzer generation

This untracked supervisor eventually calls `harbor_stop_analyzer`, which rereads the shared PID file. If `RESET_RUN` installs a replacement analyzer while the old supervisor is draining, the old supervisor can terminate the new process; the deterministic reproduction stopped `new-analyzer`. Track and stop the supervisor during reset, and capture/check the analyzer PID or generation it owns before stopping anything.

Agents/utils/common/Harbor/start.sh
```diff
@@ -143,6 +143,172 @@ harbor_start_monitor_if_enabled() {
...
+      ;;
+  esac
+  printf '%s\n' "$analyzer_pid" >"$HARBOR_ANALYZER_PID_FILE"
+  for _ in $(seq 1 50); do
+    if ! kill -0 "$analyzer_pid" >/dev/null 2>&1; then
+      rm -f "$HARBOR_ANALYZER_PID_FILE"
+      echo "[ERROR] Harbor analyzer exited during startup; see $HARBOR_ANALYZER_LOG_FILE" >&2
+      exit 1
+    fi
+    if harbor_analyzer_pid_matches_run "$analyzer_pid"; then
```
qiyuangong, line 212: [P2] Require an analyzer-owned readiness signal

A PID/cmdline match proves only that `exec` occurred, not that imports and argument validation completed. A stub child that stayed alive for 0.2 seconds and then exited 2 still returned launcher success, left the PID file present, and was gone one second later. Require a readiness file/pipe or a meaningful stability check, and clear the PID file if readiness fails.

Agents/utils/common/Harbor/start.sh
```diff
@@ -175,8 +341,17 @@ if [[ $# -gt 0 ]]; then
   if [[ "$ROLLOUT" != "1" ]]; then
     harbor_start_online_analysis_if_enabled
     harbor_start_monitor_if_enabled
+    if ! harbor_start_analyzer_if_enabled; then
+      harbor_stop_monitor >/dev/null 2>&1 || true
```
qiyuangong, line 345: [P2] Stop online analysis when analyzer startup fails

`harbor_start_online_analysis_if_enabled` has already run, but this branch stops only the Harbor monitor before exiting. The two zellij analyzer-start failure branches have the same omission. When both features are enabled and analyzer startup fails, the online rule analyzer remains running indefinitely. Stop it on every failure exit, or centralize startup rollback cleanup.

qiyuangong (commented):
Four lifecycle issues remain on the current head; each is anchored below with the affected behavior and requested fix.

PR #3, opened 2026-07-23
Migrate control-plane workflows from Claude Code to Pi

JasperChan929:
## 1. Why the change?

The repository currently requires Claude Code in the host or DinD controller to translate user prompts into FleetSpec. Replace that control-plane dependency with Pi while retaining the existing Claude Code benchmark agent inside Harbor task containers.

Closes sii-system/sii-agent-fleet#92.

## 2. Summary of the change

- Install and pin `@earendil-works/pi-coding-agent@0.81.1` for control-plane use and enforce its Node.js minimum.
- Merge the managed Pi gateway provider, settings, and repository skills without overwriting unrelated Pi or existing Claude configuration.
- Run Prompt-to-FleetSpec translation through an isolated, tool-free Pi JSONL process with lifecycle, provider-error, stop-reason, and final-output validation.
- Update DinD bootstrap checks to verify Pi in the controller while preserving Claude Code artifacts and `agent=claude-code` dispatch for nested Harbor containers.
- Add focused setup, Prompt JSONL, and DinD regression coverage and update active documentation.

## 3. Other details

The Harbor Claude adapter, FleetSpec agent semantics, Opik hooks, worker behavior, RL paths, and pinned task-container Claude Code version remain unchanged.

Validation:

- `python3 -m unittest discover -s scripts/tests` — 90 tests passed.
- Shell syntax checks, Python compilation checks, and `git diff --check`.
- A real Prompt run used Pi to generate `agent=claude-code`, launched Harbor, installed Claude Code 2.1.90 inside the task container, and completed the verifier path. The selected `compile-compcert` trial scored 0 because the inner Claude runtime responded to the appended language instruction instead of performing the task; the control-plane migration and dispatch completed successfully.

## Porting notes

Ported from sii-system/sii-agent-fleet#104 onto this repository's history. Two adaptations:

- `test_setup.py` no longer expects the retired `tui-dashboard-deployment` skill; main removed that subsystem, and `setup.sh` here installs the remaining three skills.
- The skill-symlink assertion resolves both sides, so the test also passes on macOS where the tmpdir path traverses the `/var -> /private/var` symlink.

Validation on this branch: `test_setup.py` + `test_fleet_prompt.py` 37 passed; `test_dind_run.sh` passes. The unrelated pre-existing failure in `test_run_fleet.py::test_explicit_local_taskset_maps_only_path_inputs` also fails on clean `main` and is untouched here.

### DIFF
--- a/scripts/dind-run.sh
+++ b/scripts/dind-run.sh
@@ -403,7 +403,10 @@ case "$DIND_BOOTSTRAP" in
     run_setup=1
     ;;
   missing)
-    if ! docker_exec sh -c 'command -v pi >/dev/null 2>&1 && pi --version 2>/dev/null | grep -Fqx "$3" && test -f "$1" && test -d "$2"' \
+    # Extract the version number instead of matching the whole output line:
+    # setup.sh does the same, and an exact-line match would re-run setup on
+    # every invocation if `pi --version` ever prints more than the bare version.
+    if ! docker_exec sh -c 'command -v pi >/dev/null 2>&1 && [ "$(pi --version 2>/dev/null | grep -oE "[0-9]+\.[0-9]+\.[0-9]+" | head -n 1)" = "$3" ] && test -f "$1" && test -d "$2"' \
       sh "$DIND_HOME_DIR/.pi/agent/models.json" \
       "$DIND_HOME_DIR/.pi/agent/skills/harbor-benchmark-runner" \
       "${PI_VERSION:-0.81.1}"; then
--- a/scripts/dind-run.sh
+++ b/scripts/dind-run.sh
@@ -375,7 +375,7 @@ declare -a run_env=(
   "HOME=$DIND_HOME_DIR"
 )
 run_env+=("${proxy_env[@]}")
-for optional in TRACE_TO_OPIK OPIK_URL OPIK_API_KEY OPIK_WORKSPACE OPIK_PROJECT_NAME CLAUDE_TGZ_SOURCE CLAUDE_WHEEL_DIR_SOURCE TB_CC_CLAUDE_TGZ_SOURCE TB_CC_PY_WHEEL_DIR_SOURCE; do
+for optional in PI_VERSION TRACE_TO_OPIK OPIK_URL OPIK_API_KEY OPIK_WORKSPACE OPIK_PROJECT_NAME CLAUDE_TGZ_SOURCE CLAUDE_WHEEL_DIR_SOURCE TB_CC_CLAUDE_TGZ_SOURCE TB_CC_PY_WHEEL_DIR_SOURCE; do
   if [[ -n "${!optional:-}" ]]; then
     run_env+=("$optional=${!optional}")
   fi
@@ -403,7 +403,10 @@ case "$DIND_BOOTSTRAP" in
     run_setup=1
     ;;
   missing)
-    if ! docker_exec sh -c 'command -v claude >/dev/null 2>&1 && test -d "$1"' sh "$DIND_HOME_DIR/.claude/skills/sii-agent-fleet"; then
+    if ! docker_exec sh -c 'command -v pi >/dev/null 2>&1 && pi --version 2>/dev/null | grep -Fqx "$3" && test -f "$1" && test -d "$2"' \
+      sh "$DIND_HOME_DIR/.pi/agent/models.json" \
+      "$DIND_HOME_DIR/.pi/agent/skills/harbor-benchmark-runner" \
+      "${PI_VERSION:-0.81.1}"; then
       run_setup=1
     fi
     ;;
--- a/scripts/pi_prompt.py
+++ b/scripts/pi_prompt.py
@@ -1,0 +1,323 @@
+#!/usr/bin/env python3
+"""Run the control-plane Pi translator and emit its final JSON object."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlparse
+
+
+PROVIDER = "sii-gateway"
+API_KEY_ENV = "SII_AGENT_FLEET_API_KEY"
+
+
+class PromptFailure(RuntimeError):
+    """A safe, user-facing Prompt translation failure."""
+
+
+def normalized_base_url(value: str) -> str:
+    base_url = value.strip().rstrip("/")
+    if base_url and not base_url.endswith("/v1"):
+        base_url = f"{base_url}/v1"
+    parsed = urlparse(base_url)
+    if parsed.scheme not in {"http", "https"} or not parsed.hostname:
+        raise PromptFailure("invalid BASE_URL for Pi provider")
+    return base_url
+
+
+def models_config(
+    base_url: str,
+    model: str,
+    *,
+    display_name: str = "SII Agent Fleet Prompt Translator",
+) -> dict[str, Any]:
+    """Keep this provider contract aligned with the merged Pi analyzer."""
+
+    return {
+        "providers": {
+            PROVIDER: {
+                "baseUrl": base_url,
+                "api": "openai-completions",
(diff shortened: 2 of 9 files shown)


JasperChan929:
Rebased onto main (picking up the pinned Harbor runner work from #1) and addressed the two medium review findings in 88b16f7:

- The DinD bootstrap check now extracts the version number from `pi --version` the same way setup.sh does, instead of requiring the whole output line to equal the pinned version — an exact-line mismatch would have silently re-run setup on every invocation.
- `REPO_URL` in setup.sh and the root README clone instructions now point at sii-system/agent-fleet instead of the retired sii-agent-fleet repository.

Rebase notes: setup.sh keeps main's pinned-runner step (now step 8) ahead of the Pi skills install; test_setup.py gates it off with HARBOR_RUNNER_SETUP=0 so the fixture stays hermetic; test_dind_run.sh keeps both main's stale-image tests and the Pi bootstrap tests.

Validation: test_setup.py + test_fleet_prompt.py 37 passed; test_dind_run.sh passes (run via a sed-portability shim on macOS — the GNU `sed -i` at its line 14 is a pre-existing main issue and untouched here). The remaining review suggestion — a real Pi 0.81.1 smoke (`pi --version` output format, one live prompt translation) — still stands before merge.

scripts/pi_prompt.py
```diff
@@ -0,0 +1,323 @@
...
+        root = Path(temporary)
+        runtime_dir = root / "pi-agent"
+        work_dir = root / "work"
+        runtime_dir.mkdir()
+        work_dir.mkdir()
+        (runtime_dir / "models.json").write_text(
+            json.dumps(models_config(base_url, model), indent=2) + "\n",
+            encoding="utf-8",
+        )
+        command = [
```
LongHanyu, line 249: The prompt appears to be passed to Pi through stdin instead of as a positional message. Pi 0.81.1 print mode consumes `initialMessage` / positional `messages`, and the existing Harbor analyzer path appends the prompt as the last argv element. With the current command, real Pi may run without the user request and return no final assistant message, while the unit test still passes because its fake `pi` reads stdin. Please append `args.prompt` to this command and stop passing it via `subprocess.run(input=...)`.

scripts/tests/test_fleet_prompt.py
```diff
@@ -34,26 +35,33 @@ def setUp(self):
...
-    printf 'arg=<%s>\\n' "$@"
-  } >"$CLAUDE_STUB_CAPTURE"
+if [[ "${1:-}" == "--version" ]]; then
+  printf '%s\n' '0.81.1'
+  exit 0
 fi
-printf '%s\\n' "$CLAUDE_STUB_RESPONSE"
-exit "${CLAUDE_STUB_EXIT:-0}"
+stub_dir="$(cd "$(dirname "$0")" && pwd)"
+prompt="$(cat)"
```
LongHanyu, line 47: This stub masks the real Pi invocation contract by reading the prompt from stdin. After changing `pi_prompt.py` to pass the prompt as a positional message, please update this test double to assert the prompt is present in argv instead of `cat`-ing stdin; otherwise the regression can come back without failing tests.

scripts/setup.sh
```diff
@@ -18,8 +18,8 @@ SOURCE_REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
 
 # ---- Hardcoded versions (override via env if needed) ----
 NODE_VERSION="${NODE_VERSION:-24}"
-CLAUDE_CODE_VERSION="${CLAUDE_CODE_VERSION:-2.1.90}"
-REPO_URL="${REPO_URL:-https://github.com/sii-system/sii-agent-fleet.git}"
+PI_VERSION="${PI_VERSION:-0.81.1}"
+REPO_URL="${REPO_URL:-https://github.com/sii-system/agent-fleet.git}"
 REPO_DIR="${REPO_DIR:-$HOME/sii-agent-fleet}"
```
LongHanyu, line 23: The repo URL now points at `sii-system/agent-fleet`, but the default clone directory is still `$HOME/sii-agent-fleet`. The README mostly avoids this by recommending `REPO_DIR="$PWD"`, but users who run setup directly will get a checkout under the old project name. Please either switch the default to `$HOME/agent-fleet` or document that the old path is intentional for compatibility.

scripts/pi_prompt.py
```diff
@@ -0,0 +1,323 @@
...
+    base_url = value.strip().rstrip("/")
+    if base_url and not base_url.endswith("/v1"):
+        base_url = f"{base_url}/v1"
+    parsed = urlparse(base_url)
+    if parsed.scheme not in {"http", "https"} or not parsed.hostname:
+        raise PromptFailure("invalid BASE_URL for Pi provider")
+    return base_url
+
+
+def models_config(
```
LongHanyu, line 36: This helper now duplicates a fair amount of Pi runtime logic that already exists in the Harbor analyzer path: provider config construction, minimal environment setup, JSONL event validation, provider error extraction, and final assistant message parsing. I do not think this has to block the migration, since `fleet_prompt.sh` should not depend directly on Harbor analyzer internals, but please consider extracting a shared Pi runtime module in a follow-up so the two callers do not drift when Pi changes its event schema or provider config requirements.

scripts/pi_prompt.py
```diff
@@ -0,0 +1,323 @@
...
+            PROVIDER: {
+                "baseUrl": base_url,
+                "api": "openai-completions",
+                "apiKey": f"${API_KEY_ENV}",
+                "compat": {
+                    "supportsDeveloperRole": False,
+                    "supportsReasoningEffort": False,
+                    "supportsUsageInStreaming": True,
+                    "maxTokensField": "max_tokens",
+                    "thinkingFormat": "zai",
```
qiyuangong, line 55: [P1] This turns the generic `BASE_URL`/`MODEL` contract into a ZAI-specific OpenAI Chat Completions profile. `config.env` still advertises OpenAI- or Anthropic-compatible endpoints, but `api: openai-completions` drops Anthropic `/v1/messages`, while `reasoning: true` plus `thinkingFormat: zai` makes Pi 0.81.1 send a `thinking` object to every gateway (confirmed with a local request capture). Anthropic-only and non-ZAI OpenAI-compatible gateways can therefore fail before translation. Please make the Pi API/compat profile explicit or configurable, and add coverage for the supported protocol variants.

qiyuangong (commented):
Two actionable compatibility/bootstrap gaps found on the exact PR head. This is a non-blocking follow-up because #3 is closed and superseded by #8; both issues still reproduce in the implementation currently on main.

scripts/dind-run.sh
```diff
@@ -403,7 +403,13 @@ case "$DIND_BOOTSTRAP" in
...
     ;;
   missing)
-    if ! docker_exec sh -c 'command -v claude >/dev/null 2>&1 && test -d "$1"' sh "$DIND_HOME_DIR/.claude/skills/sii-agent-fleet"; then
+    # Extract the version number instead of matching the whole output line:
+    # setup.sh does the same, and an exact-line match would re-run setup on
+    # every invocation if `pi --version` ever prints more than the bare version.
+    if ! docker_exec sh -c 'command -v pi >/dev/null 2>&1 && [ "$(pi --version 2>/dev/null | grep -oE "[0-9]+\.[0-9]+\.[0-9]+" | head -n 1)" = "$3" ] && test -f "$1" && test -d "$2"' \
+      sh "$DIND_HOME_DIR/.pi/agent/models.json" \
+      "$DIND_HOME_DIR/.pi/agent/skills/harbor-benchmark-runner" \
+      "${PI_VERSION:-0.81.1}"; then
```
qiyuangong, line 412: [P2] `DIND_BOOTSTRAP=missing` checks only `models.json` and `harbor-benchmark-runner`, although `setup.sh` installs three managed Pi skills. If either OpenClaw skill is absent, this condition succeeds and setup is skipped, leaving the controller incomplete. Please check all managed skill paths and add a regression case for each missing skill.

PR #5, opened 2026-07-23
Add an isolated Harbor wrapper for mid-turn model fusion

sgwhat:
## 1. Why the change?

  Mid-turn model fusion needs to run Claude Code with Router-generated subagents, fusion gate hooks, and task-specific mounts.

  This integration should only be enabled when explicitly running the model-fusion task script.
  
Normal Harbor tasks must keep their existing environment, command construction, and Claude Code behavior unchanged.

  ## 2. Summary of the change

  Add an isolated integration under `Agents/utils/common/Harbor/model-fusion/`.

  The new flow:

  1. Uses the sibling `sii-fusion-router` checkout to prepare the task prompt, Claude Code agent
  definitions, and fusion contract.
  2. Runs the existing Harbor worker through scoped wrappers that inject only the required fusion
  environment variables and read-only mounts.
  3. Applies the fusion-specific Claude Code runtime overlay, including the round gate,
  `--agents` configuration, and file-backed system prompt transport.
  4. Uses the Router implementation again to finalize the fusion artifact after the Harbor task
  completes.

  The shared Harbor `env.sh`, `harboropik.sh`, worker implementation, global Claude Code
  integration, skills, and tests remain unchanged.

  ## 3. Other details

  - The integration currently targets a single Terminal-Bench 2.1 task per run.
  - A sibling `sii-fusion-router` checkout is required. Missing Router builders, prompts, gate
  code, or templates produce an explicit preflight error.
  - `MID_TURN_PREPARE_ONLY=1` can be used to generate and inspect the fusion contract without
  starting Harbor.
  - Normal Harbor runs do not source or import anything from the `model-fusion` directory.


Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
```diff
@@ -0,0 +1,212 @@
...
+    if [[ -d "$candidate/$TASK_ID" ]]; then
+      DATASET_PATH="$candidate"
+      break
+    fi
+  done
+fi
+: "${DATASET_PATH:?set DATASET_PATH to a Terminal-Bench task directory root}"
+
+export RUN_ID OUTPUT_ROOT OUTPUT_PATH FUSION_ROUTER_DIR
+export DATASET_NAME="${DATASET_NAME:-terminalbench21}"
```
qiyuangong: [P1] Keep the task contract and executed dataset on the same source

This default makes the shared Harbor wrapper choose the registry path (`--dataset terminal-bench/terminal-bench-2-1`), while fusion preparation below reads and mounts the instruction from local `DATASET_PATH`. A modified or different local checkout can therefore produce a contract for one task version while Harbor evaluates another. Please default this local-checkout flow to `DATASET_NAME=auto`, or build the contract from the registry task that Harbor actually selects.

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
```diff
@@ -0,0 +1,212 @@
...
+SPAN_PANEL_MODELS="${SPAN_PANEL_MODELS:-minimax2.7,minimax2.7}"
+SPAN_OUTER_MODEL="${SPAN_OUTER_MODEL:-minimax2.7}"
+SPAN_ARTIFACT_DIR="${SPAN_ARTIFACT_DIR:-${OUTPUT_PATH}/mid-turn-fusion/${TASK_ID}}"
+SPAN_FUSION_JSON_COPY="${SPAN_FUSION_JSON_COPY:-${OUTPUT_PATH}/${TASK_ID}.fusion.json}"
+SPAN_TASK_SUBAGENT_PROMPT_FILE="${SPAN_TASK_SUBAGENT_PROMPT_FILE:-${SPAN_ARTIFACT_DIR}/task-subagent-system-prompt.md}"
+SPAN_TASK_SUBAGENT_AGENTS_FILE="${SPAN_TASK_SUBAGENT_AGENTS_FILE:-${SPAN_ARTIFACT_DIR}/claude-agents.json}"
+
+export MODEL="$MAIN_MODEL"
+export TB_MODEL="$MAIN_MODEL"
+export TB_ANTHROPIC_MODEL="${TB_ANTHROPIC_MODEL:-$MAIN_MODEL}"
```
qiyuangong: [P2] Apply `MAIN_MODEL` to the already-derived Anthropic aliases

The shared `env.sh` has already populated `TB_ANTHROPIC_MODEL` and the default-model aliases from the original `MODEL`, so these `:-$MAIN_MODEL` assignments retain the old values. I reproduced `MODEL=new-model` and `TB_MODEL=new-model` with `TB_ANTHROPIC_MODEL=old-model`; Harbor passes the latter through `--ae`, where it wins in the task environment. Preserve caller-explicit aliases before sourcing, then overwrite only the derived defaults when `MAIN_MODEL` is selected.

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
```diff
@@ -0,0 +1,212 @@
...
+
+if [[ "$MID_TURN_PREPARE_ONLY" == "1" ]]; then
+  printf 'prepared mid-turn fusion contract at %s\n' "$SPAN_ARTIFACT_DIR"
+  exit 0
+fi
+
+harbor_init_run_dirs
+harbor_reset_run_state
+harbor_ensure_dataset
+harbor_prepare_task_file
```
qiyuangong, line 234: [P1] Replace the task file when resetting this single-task run

`harbor_reset_run_state` does not remove `TASK_FILE`, and `harbor_prepare_task_file` regenerates it only when `RESET_RUN=1` or the file is absent. Reusing an output path therefore keeps the previous task; I reproduced a reset after changing the source from `first-task` to `second-task` and the prepared task remained `first-task`. This is reachable by default because `RUN_ID` has minute granularity and excludes `TASK_ID`, and retained jobs can also be selected during finalization. Please make the run ID unique and explicitly refresh the task/job state for this wrapper.

qiyuangong (commented):
Three configuration and run-state issues can make the fusion artifact disagree with the benchmark that actually ran. These are non-blocking review comments, but I recommend addressing them before relying on the resulting benchmark data.

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
sgwhat: Fixed in latest commit

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
```diff
@@ -0,0 +1,250 @@
...
+  echo "[ERROR] only SPAN_FORCE_MODE=mid-turn-fusion is supported; got: $SPAN_FORCE_MODE" >&2
+  exit 2
+fi
+
+if [[ -z "${DATASET_PATH:-}" ]]; then
+  for candidate in \
+    /home/tianyi/terminal-bench-2-verified \
+    /home/siisyslab/tianyi/terminal-bench-2-verified \
+    /workspace/terminal-bench-2-verified \
+    /workspace/terminal-bench-2-1/tasks; do
```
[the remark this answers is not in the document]
sgwhat: Fixed in 61343a1: removed the /home/tianyi and /home/siisyslab fallbacks. Auto-detection now only checks portable /workspace dataset roots; callers can still set DATASET_PATH explicitly.

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
```diff
@@ -0,0 +1,250 @@
...
+  TB_CLAUDE_CODE_SUBAGENT_MODEL="$_USER_SUBAGENT_MODEL_VALUE"
+else
+  TB_CLAUDE_CODE_SUBAGENT_MODEL=""
+fi
+
+MAIN_MODEL="${MAIN_MODEL:-$MODEL}"
+SPAN_PANEL_MODELS="${SPAN_PANEL_MODELS:-minimax2.7,minimax2.7}"
+SPAN_OUTER_MODEL="${SPAN_OUTER_MODEL:-minimax2.7}"
+SPAN_ARTIFACT_DIR="${SPAN_ARTIFACT_DIR:-${OUTPUT_PATH}/mid-turn-fusion/${TASK_ID}}"
+SPAN_FUSION_JSON_COPY="${SPAN_FUSION_JSON_COPY:-${OUTPUT_PATH}/${TASK_ID}.fusion.json}"
```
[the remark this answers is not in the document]
sgwhat: Fixed in 61343a1: SPAN_ARTIFACT_DIR and SPAN_FUSION_JSON_COPY now use task_slug, while TASK_ID remains unchanged for dataset lookup and task selection. Added test_task_id_is_sanitized_only_for_output_paths covering TASK_ID=nested/task.

Agents/utils/common/Harbor/model-fusion/harboropik.sh
```diff
@@ -0,0 +1,92 @@
...
+  python3 "$MODEL_FUSION_DIR/harbor_worker_utils.py" \
+    append-readonly-mounts "$mounts_json" \
+    --mount "$TB_FUSION_ROUND_ROUTER_DIR" "$TB_FUSION_ROUND_ROUTER_MOUNT_PATH" \
+    --mount "$TB_FUSION_TASK_FILE_SOURCE" "$TB_FUSION_TASK_FILE"
+)"
+if ((mount_value_index >= 0)); then
+  args[$mount_value_index]="$mounts_json"
+else
+  args+=(--mounts-json "$mounts_json")
+fi
```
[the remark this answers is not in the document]
sgwhat: Fixed in 61343a1: added test_model_fusion_harboropik.py covering existing mounts preservation, Router/task mount injection, mounts creation when absent, all injected --ae values, and transparent --help forwarding. Full Harbor Python suite: 34 passed.

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
sgwhat: Fixed in 8475797: DATASET_NAME now defaults to auto so Harbor executes the same local DATASET_PATH used to build the fusion task contract. Covered by test_defaults_use_local_dataset_and_unique_task_run_ids.

Agents/utils/common/Harbor/model-fusion/run_one_tb21_task.sh
sgwhat: Fixed in 8475797: caller-explicit Anthropic aliases are snapshotted before env.sh is sourced; only shared derived defaults are replaced by MAIN_MODEL. Covered by the derived-alias and explicit-alias regression tests.

qiyuangong (approved):
LGTM

Issue #7, opened 2026-07-23
Extract a shared Pi runtime module for the control-plane translator and Harbor analyzer


JasperChan929:
Follow-up from the review of the Pi control-plane migration (#3): `scripts/pi_prompt.py` duplicates Pi runtime logic that also lives in `Agents/utils/common/Harbor/scripts/harbor_analyzer/pi.py` — provider config construction, minimal environment setup, JSONL event validation, provider-error extraction, and final assistant message parsing. The two callers can drift when Pi changes its event schema or provider config requirements.

One constraint shapes the design: the Harbor tree is packaged into the runner image and needs to stay self-contained, so a naive cross-tree import from `scripts/` would break inside containers. Options include a build-time vendored copy with a sync check, or moving the shared module into a location both packaging paths include.

PR #8, opened 2026-07-23
Migrate control-plane workflows from Claude Code to Pi

JasperChan929:
## 1. Why the change?

The repository currently requires Claude Code in the host or DinD controller to translate user prompts into FleetSpec. Replace that control-plane dependency with Pi while retaining the existing Claude Code benchmark agent inside Harbor task containers.

Supersedes #3, resubmitted from a fork branch as required by the contribution workflow. Closes sii-system/sii-agent-fleet#92.

## 2. Summary of the change

- Install and pin `@earendil-works/pi-coding-agent@0.81.1` for control-plane use and enforce its Node.js minimum.
- Merge the managed Pi gateway provider, settings, and repository skills without overwriting unrelated Pi or existing Claude configuration.
- Run Prompt-to-FleetSpec translation through an isolated, tool-free Pi JSONL process with lifecycle, provider-error, stop-reason, and final-output validation.
- Update DinD bootstrap checks to verify Pi in the controller while preserving Claude Code artifacts and `agent=claude-code` dispatch for nested Harbor containers.
- Add focused setup, Prompt JSONL, and DinD regression coverage and update active documentation.

## 3. Review follow-up from #3

All four review comments from @LongHanyu on #3 are addressed in the head commit:

- The Prompt translation is now passed to Pi as the trailing positional message with stdin closed, matching the Harbor analyzer invocation contract, instead of being written to stdin.
- The `pi` test stub records argv and stdin separately, and the regression test asserts the prompt arrives as the final positional argument with nothing on stdin.
- `setup.sh` now defaults `REPO_DIR` to `$HOME/agent-fleet` to match the renamed repository; the root README's mention of the default is updated too.
- Extracting a shared Pi runtime module for `pi_prompt.py` and the Harbor analyzer is tracked in #7 as a follow-up, as suggested.

Additionally, `scripts/tests/test_dind_run.sh` no longer uses GNU-only `sed -i`/`0,/re/` addressing, so the DinD regression suite also runs on macOS (BSD sed).

## 4. Other details

The Harbor Claude adapter, FleetSpec agent semantics, Opik hooks, worker behavior, RL paths, and pinned task-container Claude Code version remain unchanged.

Validation on this branch:

- `python3 -m unittest discover -s scripts/tests` — 90 tests, 1 pre-existing failure (`test_run_fleet.py::test_explicit_local_taskset_maps_only_path_inputs`, also fails on clean `main` on macOS; untouched here).
- `bash scripts/tests/test_dind_run.sh` — passes, including on macOS after the portability fix.
- Shell syntax checks, Python compilation checks, and `git diff --check`.
- A real Prompt run previously used Pi to generate `agent=claude-code`, launched Harbor, installed Claude Code 2.1.90 inside the task container, and completed the verifier path.

## Porting notes

Ported from sii-system/sii-agent-fleet#104 onto this repository's history. Two adaptations:

- `test_setup.py` no longer expects the retired `tui-dashboard-deployment` skill; main removed that subsystem, and `setup.sh` here installs the remaining three skills.
- The skill-symlink assertion resolves both sides, so the test also passes on macOS where the tmpdir path traverses the `/var -> /private/var` symlink.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ Claude Code.
 ### 2. Clone the repository
 
 ```bash
-git clone --recurse-submodules https://github.com/sii-system/sii-agent-fleet.git
-cd sii-agent-fleet
+git clone --recurse-submodules https://github.com/sii-system/agent-fleet.git
+cd agent-fleet
 ```
 
 ### 3. Configure and set up
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ REPO_DIR="$PWD" ./scripts/setup.sh
 ```
 
 `REPO_DIR="$PWD"` points setup at this checkout instead of its default
-`$HOME/sii-agent-fleet` clone.
+`$HOME/agent-fleet` clone.
 
 ### 4. Run one benchmark
 
--- a/README.md
+++ b/README.md
@@ -13,7 +13,9 @@ evaluating Claude Code, OpenCode, and OpenClaw.
 
 Install tmux with your package manager and install Zellij from its
 [releases page](https://github.com/zellij-org/zellij/releases). `setup.sh`
-below installs Node.js and Claude Code for you.
+below installs Node.js and Pi for the control-plane Prompt workflow. Harbor
+task containers still install and run the selected benchmark agent, including
+Claude Code.
 
 ### 2. Clone the repository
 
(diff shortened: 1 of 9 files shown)


LongHanyu (approved):
LGTM

PR #10, opened 2026-07-23
Fix bare-metal setup and Harbor run reporting

JasperChan929:
## Why the change

Bare-metal setup could report success while later runs still failed because managed tools were missing from `PATH`, Pi tried to install under `/usr/lib`, or output defaulted to unwritable `/workspace/runs`. The initial fix also risked replacing user-owned tools in `~/.local/bin`. When a run did start, Zellij could close without clearly reporting completion, reward, result paths, or the real Harbor exit code.

Addresses https://github.com/sii-system/sii-agent-fleet/issues/105.

## Summary of the change

- Validate host requirements and provision pinned Zellij 0.44.3, uv/uvx 0.11.28, and Pi 0.81.1 in project-private, user-writable locations.
- Verify upstream release checksums, cache downloads, install atomically, and preserve existing user-owned executables.
- Persist a non-secret path contract loaded by public runners, so setup works immediately without sourcing `~/.bashrc`.
- Use a secure user-owned Zellij runtime directory and default host outputs and dependency caches to repository/XDG paths.
- Reuse existing local configuration, prompt only for missing values, make Opik tracing opt-in, and fail fast when required run configuration is missing.
- Generate bounded session/project names from the effective run and print the run ID, session, output, summary, and attach command at startup.
- Auto-close Zellij only after confirmed success; foreground runs print final counts and reward and return the recorded Harbor exit code. Failures retain actionable diagnostics.
- Sync the renamed tracing-plugin submodule, require Docker daemon access before setup reports success, and keep local configuration and backups private.

## Other details

This is a bare-metal host setup and run-usability change. DinD runner behavior is unchanged. It does not replace or reconfigure Claude Code inside benchmark task containers.

Setup still expects Docker Engine, Docker Compose v2, Python >=3.9, git, curl, jq, openssl, util-linux, and procps to be installed by the host administrator.

Validation:

- 99 script/router/setup Python tests and 29 Harbor Python tests passed.
- All affected prerequisite, Harbor, and OpenClaw shell tests passed.
- Fresh-HOME setup installed Zellij, uv/uvx, Pi, and the pinned Harbor runner; a clean shell resolved the managed commands without manual `PATH` changes.
- A real Pi Prompt dry-run produced the expected TerminalBench FleetSpec.
- A real foreground `fix-git` canary completed 1/1 with reward 1 and zero errors, printed the summary, and returned exit code 0.


JasperChan929:
Checksum verification note: Zellij v0.44.3 publishes a checksum for the extracted binary, not for the `.tar.gz` archive. The official [`zellij-x86_64-unknown-linux-musl.sha256sum`](https://github.com/zellij-org/zellij/releases/download/v0.44.3/zellij-x86_64-unknown-linux-musl.sha256sum) contains:

```text
397481870c4fc3bae646cd7613cde3a1cebdc204558a6cb9a7c603d4c852fc90  target/x86_64-unknown-linux-musl/release/zellij
```

Accordingly, the `binary` checksum mode in `scripts/prerequisites.sh` is intentional: it extracts the archive and verifies the `zellij` executable. Using archive mode would compare the binary digest against the tarball and fail. `scripts/tests/test_prerequisites.sh` also covers this distinction with a binary checksum fixture for Zellij and an archive checksum fixture for uv. The rationale is now documented inline in commit `a51be54`.

qiyuangong (commented):
Reviewed current head `2c273989a452238841ee7aa3effd4743394f24fa`. Five functional regressions remain; details are attached to the affected lines.

scripts/prerequisites.sh
```diff
@@ -0,0 +1,381 @@
...
+  local temporary="${AGENT_FLEET_PATHS_FILE}.tmp.$$"
+  mkdir -p "$(dirname "$AGENT_FLEET_PATHS_FILE")" ||
+    { agent_fleet_prereq_error "cannot create paths-file directory: $AGENT_FLEET_PATHS_FILE"; return 1; }
+  {
+    printf '# Managed by agent-fleet scripts/prerequisites.sh\n'
+    printf 'export AGENT_FLEET_BIN_DIR=%q\n' "$AGENT_FLEET_BIN_DIR"
+    printf 'export AGENT_FLEET_CACHE_DIR=%q\n' "$AGENT_FLEET_CACHE_DIR"
+    printf 'export AGENT_FLEET_NODE_BIN_DIR=%q\n' "$AGENT_FLEET_NODE_BIN_DIR"
+    printf 'export AGENT_FLEET_NPM_BIN_DIR=%q\n' "$AGENT_FLEET_NPM_BIN_DIR"
+    printf 'export AGENT_FLEET_RUNTIME_DIR=%q\n' "$AGENT_FLEET_RUNTIME_DIR"
```
qiyuangong: [P1] This persists a UID-expanded runtime path from setup. In the supported DinD flow, setup runs as root with `HOME=/home/agent`, so `paths.env` records `/tmp/agent-fleet-z-0`; the benchmark then runs as `DIND_USER` and rejects that directory as foreign-owned. I reproduced the exact failure in a disposable container. Run setup as `DIND_USER`, or avoid persisting this UID-derived fallback and initialize it after switching users.

Agents/utils/common/Harbor/run_harbor_registry.sh
```diff
@@ -9,13 +9,36 @@ set +e
...
+  status=1
+  printf '%s\n' "$status" > "$HARBOR_BENCHMARK_EXIT_FILE"
+fi
+
+if [[ "$status" -ne 0 ]]; then
+  show_registry_summary
+  echo
+  echo "Harbor failed; keeping this pane open for diagnostics."
+  echo "Press Ctrl-q to leave Zellij after reviewing the error above."
+  while true; do
```
qiyuangong: [P1] Every nonzero Harbor result enters this infinite loop, including the default foreground and noninteractive paths. Those callers never reach `start.sh`'s summary/exit-code reporting without a manual Ctrl-q, so automated runs hang forever. Retain the pane only for detached or explicitly interactive runs; otherwise return the recorded failure.

scripts/setup.sh
```diff
@@ -15,17 +15,147 @@ err()   { echo -e "\033[1;31m[FAIL]\033[0m  $*" >&2; }
...
+      ;;
+    *)
+      return 1
+      ;;
+  esac
+}
+
+prompt_trace_to_opik() {
+  local reply
+  if [[ -n "${TRACE_TO_OPIK:-}" ]]; then
```
qiyuangong, line 111: [P1] In DinD this value is not necessarily an explicit user choice: `dind-run.sh` sources the committed `config.env` (`TRACE_TO_OPIK=true`) and forwards it to setup. `docker_exec_root_env` has no interactive stdin, so a fresh no-Opik bootstrap reaches the `OPIK_URL` prompt, gets EOF, and fails instead of using the advertised default-off behavior. Do not treat the template default as an explicit setup override, or make the committed default false.

scripts/setup.sh
```diff
@@ -137,11 +253,18 @@ if ! command -v npm >/dev/null 2>&1; then
   exit 1
 fi
 
+# Keep both the selected Node runtime and the managed Pi installation
+# discoverable without sourcing an interactive shell startup file.
+AGENT_FLEET_NODE_BIN_DIR="$(dirname "$(command -v node)")"
+AGENT_FLEET_NPM_PREFIX="${AGENT_FLEET_NPM_PREFIX:-$AGENT_FLEET_CACHE_DIR/npm}"
```
qiyuangong, line 259: [P2] Moving Pi into this private prefix makes the existing `DIND_BOOTSTRAP=missing` probe stale: it still runs `command -v pi` under a plain shell without loading `paths.env`, so it misses the installed binary and reruns setup every invocation. I reproduced `probe=miss`. Update the probe to source the saved path contract or test this exact managed executable.

scripts/run_fleet.sh
```diff
@@ -47,6 +50,54 @@ run_command() {
...
+  done
+}
+
+validate_run_config() {
+  local -a missing=()
+  local required
+
+  API_KEY="${API_KEY:-${AUTH_TOKEN:-}}"
+  export API_KEY
+  for required in BASE_URL API_KEY MODEL; do
```
qiyuangong, line 103: [P2] This guard is stricter than Harbor's downstream resolution. `env.sh` still accepts `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, and `TB_MODEL`, but `run_fleet.sh` rejects a run configured only with those supported aliases before Harbor starts. Normalize the same aliases here before validating; the alias-only invocation currently reports all three canonical values missing.

README.md
```diff
@@ -9,11 +9,13 @@ evaluating Claude Code, OpenCode, and OpenClaw.
 
 - Docker with Docker Compose v2
 - Python 3.9 or newer
-- `git`, `curl`, `jq`, `openssl`, `tmux`, and `zellij`
+- `git`, `curl`, `jq`, and `openssl`
+- Linux `util-linux` (`flock`, `setsid`, and `script`) and `procps` (`pkill`)
```
qiyuangong, line 13: If these packages are default installed, then remove them from doc.

README.md
```diff
@@ -9,11 +9,13 @@ evaluating Claude Code, OpenCode, and OpenClaw.
...
 
-Install tmux with your package manager and install Zellij from its
-[releases page](https://github.com/zellij-org/zellij/releases). `setup.sh`
-below installs Node.js and Pi for the control-plane Prompt workflow. Harbor
+These system commands must be on `PATH`; downloading an executable without
```
qiyuangong, line 15: Don't mention too much details  in ./README.md. Move to scripts/README.MD

README.md
```diff
@@ -64,7 +69,12 @@ Then start the full benchmark, with direct arguments or in natural language
 ```
 
 The first run is slower while Harbor downloads the taskset and Docker images.
-Rerun `setup.sh` only when configuration changes.
+At launch, the runner prints the `RUN_ID`, Zellij session, output directory,
```
qiyuangong, line 72: User don't need to know so much details.

qiyuangong (approved):
LGTM

PR #11, opened 2026-07-23
Fix Harbor analyzer runtime model configuration

0407nancy:
## Why the change?

The analyzer runtime should use the model configured by the user instead of silently defaulting to a GLM model. It should also fail fast when no analyzer model is configured, and custom OpenAI-compatible providers need bearer auth enabled.

This PR migrates sii-system/sii-agent-fleet#102 to sii-system/agent-fleet.

## Summary of the change

- Remove hard-coded GLM model defaults from analyzer runtime configuration.
- Return a clear pi_model_not_configured block reason when no model is provided.
- Add authHeader to the analyzer Pi provider config.
- Update Pi analyzer module descriptions to avoid implying a fixed model.

## Validation

- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test*.py'`\n- `git diff --cached --check`


qiyuangong (approved):
LGTM

PR #14, opened 2026-07-23
Add self-hosted PR quick checks

qiyuangong:
## 1. Why the change?

The repository now has a persistent self-hosted runner but no fast validation workflow for pull requests. Contributors need early feedback on shell syntax and the lightweight unit suites before merge.

## 2. Summary of the change

- Add a `PR Quick CI` workflow for non-draft pull requests from trusted branches in this repository.
- Run the job on the self-hosted Linux x64 runner with read-only repository permissions, a pinned checkout action, no persisted credentials, cancellation of superseded runs, and a 10-minute timeout.
- Validate every tracked shell script and run the root, control-plane, DinD, and GitHub automation test suites.
- Skip fork pull requests so untrusted code cannot execute on the persistent runner attached to this public repository.
- Add contract tests that preserve the trigger, runner, least-privilege, and test-suite requirements.

### Test plan / Verification

- `docker run --rm --volume "$PWD:/repo" --workdir /repo rhysd/actionlint:1.7.12 -color`
- Parse `.github/workflows/pr-ci.yml` with PyYAML.
- `git ls-files -z '*.sh' | xargs -0 -r -n1 bash -n`
- `python3 -m unittest discover -s tests -v`
- `python3 -m unittest discover -s scripts/tests -v`
- `python3 -m unittest discover -s .github/scripts/tests -v`
- `git diff --check`

The quick checks complete in about 17 seconds on the registered host. Fork PRs remain excluded because persistent self-hosted runners can be compromised by untrusted PR code: https://docs.github.com/en/actions/reference/security/secure-use

### DIFF
--- a/.github/scripts/tests/test_pr_ci_workflow.py
+++ b/.github/scripts/tests/test_pr_ci_workflow.py
@@ -42,6 +42,7 @@ class PrCiWorkflowContractTest(unittest.TestCase):
                     f"python3 -m unittest discover -s {suite} -v",
                     self.workflow,
                 )
+        self.assertIn("bash scripts/tests/test_dind_run.sh", self.workflow)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


LongHanyu (approved):
LGTM

PR #16, opened 2026-07-24
Add self-hosted LLM reviews for drafts and updates

qiyuangong:
## 1. Why the change?

Use the higher-budget self-deployed LLM endpoint to review every PR head, including draft PRs and every commit update, without changing the existing hosted reviewer policy.

## 2. Summary of the change

- Add a separate `self-hosted-llm-pr-review.yml` workflow for opened, reopened, ready-for-review, and `synchronize` events, with drafts included.
- Keep `llm-pr-review.yml` separate because the hosted and self-hosted reviewers use different runners, environments, models, and draft policies.
- Reuse the existing review code, flow, prompt, permissions, trusted base checkout, timeout, and publication behavior through `reusable-llm-pr-review.yml`.
- Run the new reviewer on `[self-hosted, Linux, X64]` with the `self-hosted-env` GitHub Environment. Its `LLM_API_KEY` secret and `LLM_BASE_URL`/`LLM_MODEL` variables are mapped into the shared review interface.
- Give each reviewer a distinct review marker so one reviewer never suppresses the other for the same PR head.
- Preserve the merged reasoning-model handling: a 12k response budget, strict malformed-response errors, and explicit partial coverage for empty model output.
- Add regression coverage for separate workflow policy, independent reviewers, and draft review behavior.

## 3. Other details

Verified locally after merging the latest `main`:

- `python3 -m unittest discover -s .github/scripts/tests -v` (47 passed)
- `python3 -m unittest discover -s tests -q` (4 passed)
- `actionlint` 1.7.12
- Python compilation and `git diff --check`

The workflow uses `pull_request_target` and checks out only the trusted base revision; PR code is analyzed as API-provided diff text and is not executed. Because GitHub loads `pull_request_target` workflow definitions from the base branch, the self-hosted reviewer becomes active for subsequent PR events after this change reaches `main`.


.github/workflows/self-hosted-llm-pr-review.yml
```diff
@@ -0,0 +1,38 @@
+name: Self-Hosted LLM PR Review
```
LongHanyu, line 1: This workflow is almost a copy of the existing hosted LLM reviewer workflow: the differences are mainly runner, environment, credential variable names, and `LLM_REVIEW_ID`. That is acceptable if we only ever keep two reviewers, but it creates policy drift risk as the review flow evolves. Consider using a reusable workflow or colocating the hosted and self-hosted jobs in one workflow so checkout policy, permissions, timeout, and invocation stay aligned.

.github/workflows/self-hosted-llm-pr-review.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Review pull request
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          LLM_REVIEW_API_KEY: ${{ secrets.LLM_API_KEY }}
+          LLM_REVIEW_BASE_URL: ${{ vars.LLM_BASE_URL }}
+          LLM_REVIEW_MODEL: ${{ vars.LLM_MODEL }}
+          LLM_REVIEW_ID: self-hosted-llm-pr-review
+        run: >-
+          python3 .github/scripts/llm_pr_review.py
```
LongHanyu: This new self-hosted reviewer invokes the same review runtime from this branch, but that runtime still has the old 4k response budget and old content parsing behavior. If this PR lands before the response-handling fix, the self-hosted reviewer can still hit the reasoning-model empty-content failure that #15 is addressing. Please rebase this PR on the fixed runtime, or include that runtime fix here so merge order cannot leave the new reviewer broken.

.github/scripts/llm_pr_review.py
```diff
@@ -407,7 +412,8 @@ def build_summary(
     rejected: int,
     skipped: list[tuple[str, str]],
     truncated: bool,
-    incomplete_chunks: int,
+    incomplete_chunks: int = 0,
```
LongHanyu, line 415: Minor/non-blocking: when one or more chunks are incomplete, the summary can still start with “Automated review found no actionable findings.” and only later explain `Coverage: Partial` plus the empty-response note. That may be acceptable for an advisory review, but if users or branch protection read the headline as the outcome, it can be slightly misleading. Consider changing the headline for incomplete reviews to something like “Automated review was incomplete” or failing the job when `incomplete_chunks > 0`; leaving it as partial coverage is also reasonable if that is the intended policy.

LongHanyu (approved):
LGTM

PR #17, opened 2026-07-24
Start Harbor analyzer from run lifecycle

0407nancy:
## Why the change?

Harbor monitor already emits analyzer handovers, and the analyzer already supports follow mode, but they are not started under one run lifecycle. Users must launch the analyzer manually, which makes PID/log/reset behavior inconsistent with the monitor.

This PR migrates sii-system/sii-agent-fleet#95 to sii-system/agent-fleet and supersedes #2, whose original head repository is no longer connected to the target repository as a GitHub fork.

## Summary of the change

- Add Harbor analyzer lifecycle settings in `env.sh`, enabled by default.
- Start the existing `analyzer_subagent.py --follow` after the monitor produces `analyzer-handover-latest.json`.
- Add run-scoped analyzer PID matching and reset cleanup.
- Drain analyzer handovers before shutdown in command and foreground zellij modes.
- Add a detached analyzer supervisor for detached zellij runs.
- Document the analyzer startup path and `HARBOR_ANALYZER_ENABLED=0` opt-out.
- Size the drain timeout from pending handovers, analyzer timeout, retry policy, and concurrency.
- Track supervisor/analyzer PID ownership and wait for an explicit analyzer readiness file.
- Roll back online-analysis state when analyzer startup fails.

## Other details

The analyzer is enabled by default and may consume model calls. Set `HARBOR_ANALYZER_ENABLED=0` to disable it. The monitor still owns benchmark control decisions; the analyzer only consumes monitor handovers and writes reports.

The final lifecycle items address the unresolved review comments from #2/#17 with minimal changes.

Validation:
- `bash -n Agents/utils/common/Harbor/env.sh`
- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p "test*.py"` (30 tests passed)
- `git diff --check`


Agents/utils/common/Harbor/start.sh
```diff
@@ -143,6 +143,244 @@ harbor_start_monitor_if_enabled() {
...
+    processed=processed,
+    failed=failed,
+    now=time.time(),
+)
+waves = 0
+for handover, _path, _key in pending:
+    tasks = handover.get("tasks")
+    if isinstance(tasks, list) and tasks:
+        waves += math.ceil(len(tasks) / max_concurrency)
+print(max(timeout, waves * timeout))
```
qiyuangong: [P1] Include permitted retries in the drain budget

This counts one `HARBOR_ANALYZER_TIMEOUT` per task wave, but `_run_task_analysis` permits two attempts and increases a retryable 900-second dispatch timeout to 1,800 seconds. A valid task can therefore run for 2,700 seconds while the supervisor stops it after 900 seconds. Derive the deadline from the full retry policy (or wait on active analyzer work), and add a retry-path lifecycle test.

Agents/utils/common/Harbor/env.sh
```diff
@@ -495,18 +516,70 @@ harbor_stop_monitor() {
...
 
+harbor_stop_analyzer_supervisor() {
+  [[ -f "$HARBOR_ANALYZER_SUPERVISOR_PID_FILE" ]] || return 0
+  local pid
+  pid="$(cat "$HARBOR_ANALYZER_SUPERVISOR_PID_FILE" 2>/dev/null || true)"
+  if [[ ! "$pid" =~ ^[0-9]+$ ]] || ! kill -0 "$pid" >/dev/null 2>&1; then
+    rm -f "$HARBOR_ANALYZER_SUPERVISOR_PID_FILE"
+    return 0
+  fi
+  kill -TERM "$pid" >/dev/null 2>&1 || true
```
qiyuangong, line 646: [P1] Verify supervisor ownership before signaling the PID

Unlike analyzer and monitor cleanup, this sends TERM/KILL to any live PID found in the supervisor file. The supervisor does not remove its PID file on exit, so after PID reuse `RESET_RUN=1` can terminate an unrelated process; I reproduced that behavior with an unrelated sleep process. Store and check a run-specific identity, such as PID plus start time and analyzer generation, before signaling.

Agents/utils/common/Harbor/start.sh
```diff
@@ -143,6 +143,244 @@ harbor_start_monitor_if_enabled() {
...
+  (( configured_timeout > derived_timeout )) && printf '%s\n' "$configured_timeout" || printf '%s\n' "$derived_timeout"
+}
+
+harbor_start_detached_analyzer_supervisor_if_enabled() {
+  [[ "$HARBOR_ANALYZER_ENABLED" == "1" && "$ROLLOUT" != "1" ]] || return 0
+  [[ -f "$HARBOR_ANALYZER_PID_FILE" ]] || return 0
+  local analyzer_pid
+  analyzer_pid="$(cat "$HARBOR_ANALYZER_PID_FILE" 2>/dev/null || true)"
+  [[ "$analyzer_pid" =~ ^[0-9]+$ ]] || return 0
+  rm -f "$HARBOR_ANALYZER_SUPERVISOR_PID_FILE"
```
qiyuangong: [P2] Reuse or stop an existing detached supervisor

A repeated `start.sh --detach` can reuse the existing monitor and analyzer, but this unconditionally drops the prior supervisor PID and launches another. I reproduced two live supervisors with only the second PID remaining tracked, so reset cannot stop the first. Check the existing supervisor identity under a lock and return or stop it before replacement.

qiyuangong (commented):
Three lifecycle issues remain on the current head; each is anchored below with the reproduced behavior and requested fix.

Agents/utils/common/Harbor/env.sh
```diff
@@ -107,6 +107,16 @@ HARBOR_ANALYZER_BASE_URL="${HARBOR_ANALYZER_BASE_URL:-${BASE_URL:+${BASE_URL}/v1
 HARBOR_ANALYZER_MODEL="${HARBOR_ANALYZER_MODEL:-$MODEL}"
 HARBOR_ANALYZER_PI_PROVIDER="${HARBOR_ANALYZER_PI_PROVIDER:-harbor-analyzer}"
 HARBOR_ANALYZER_NO_PROXY="${HARBOR_ANALYZER_NO_PROXY:-0}"
+HARBOR_ANALYZER_ENABLED="${HARBOR_ANALYZER_ENABLED:-1}"
```
qiyuangong: [P2] Preserve the documented monitor-disabled mode

Because the analyzer now defaults to enabled, `HARBOR_MONITOR_ENABLED=0` makes `start.sh` fail unless users also set `HARBOR_ANALYZER_ENABLED=0`. I reproduced exit 1 with the documented monitor opt-out and exit 0 only after adding the analyzer opt-out. Make the analyzer default conditional on the monitor, or skip analyzer startup when monitoring is disabled.

Agents/utils/common/Harbor/scripts/analyzer_subagent.py
```diff
@@ -230,16 +232,48 @@ def _pending_handovers(
...
+        failed=failed,
+        now=now,
+    )
+    task_budget = task_analysis_timeout_budget_seconds(timeout_seconds)
+    waves = 0
+    for handover, _path, _key in pending:
+        tasks = handover.get("tasks")
+        if isinstance(tasks, list) and tasks:
+            waves += math.ceil(len(tasks) / max(1, max_concurrency))
+    return max(timeout_seconds, waves * task_budget)
```
qiyuangong: [P1] Include follow-level retries in the drain budget

This budgets one `run_handover` execution, but follow mode retries a failed handover up to `FOLLOW_MAX_FAILURE_ATTEMPTS=3`. With the default timeout, one pending task gets a 2,700-second drain budget while the full retry window is 8,100 seconds plus backoff, so teardown can stop the analyzer during its second attempt. Include the remaining follow attempts and backoff from the persisted failure state.

Agents/utils/common/Harbor/start.sh
```diff
@@ -143,6 +143,267 @@ harbor_start_monitor_if_enabled() {
...
+    fi
+  fi
+  (
+    trap '' HUP
+    while harbor_monitor_is_running_for_run; do
+      sleep 5
+    done
+    harbor_wait_for_analyzer_drain "$analyzer_pid"
+    harbor_stop_analyzer "$analyzer_pid" || true
+  ) >>"$HARBOR_ANALYZER_LOG_FILE" 2>&1 &
```
qiyuangong, line 442: [P1] Release the startup lock in the detached supervisor

This background subshell inherits fd 9 and keeps the `flock` held after the startup function returns. I reproduced the lock remaining unavailable while the supervisor was live; a second startup then blocks at `flock 9` instead of reaching the existing-supervisor check. Close fd 9 in the detached child (for example, `9>&-`) or release the lock before spawning it.

qiyuangong (commented):
Fresh review of `dc3dfbc`: three current-head lifecycle findings remain, each anchored below with reproduced behavior.

qiyuangong:
@codex review

qiyuangong:
@codex revie

qiyuangong:
@codex review

qiyuangong (approved):
LGTM

PR #18, opened 2026-07-24
Sandbox draft and fork PR checks on the self-hosted runner

qiyuangong:
## 1. Why the change?

`PR Quick CI` currently skips draft pull requests and every pull request opened from a fork. The self-hosted runner should validate both cases without directly executing untrusted candidate code in its host environment.

## 2. Summary of the change

- Trigger the controller with `pull_request_target`, including `converted_to_draft`, so draft and fork PR activity is covered.
- Keep the host-side workflow and sandbox Dockerfile pinned to the trusted base revision.
- Check out the candidate SHA without persisted credentials, but execute its files only inside a dedicated Docker container.
- Lock the container down with no network, a read-only root filesystem, no Linux capabilities, `no-new-privileges`, a non-root UID, no Docker socket, and CPU, memory, and process limits.
- Mount candidate source read-only, copy it into an ephemeral tmpfs, run shell validation and all quick test suites there, then remove per-run checkout directories and the tagged sandbox image.
- Pin the Python sandbox base image by digest and add contract tests for the event, trust boundary, checkout settings, and isolation controls.

### Test plan / Verification

- `docker run --rm --volume "$PWD:/repo" --workdir /repo rhysd/actionlint:1.7.12 -color`
- `docker build --check --file .github/ci/Dockerfile .github/ci`
- Build `.github/ci/Dockerfile` locally.
- Run the complete quick-check sequence in the container with the same network, filesystem, capability, UID, resource, tmpfs, and read-only bind-mount restrictions used by the workflow.
- `python3 -m unittest discover -s .github/scripts/tests -v`
- `git diff --check`

The isolated test sequence completes in about 13 seconds on `cpu-nat-093` after the sandbox image is available.

### Residual risk

Container isolation substantially reduces exposure but does not make a persistent self-hosted runner equivalent to an ephemeral GitHub-hosted runner; a container-runtime or kernel escape remains possible. GitHub explicitly warns that untrusted fork code on self-hosted infrastructure is risky: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories

### DIFF
--- a/.github/ci/Dockerfile
+++ b/.github/ci/Dockerfile
@@ -1,0 +1,5 @@
+FROM python:3.11-slim@sha256:233de06753d30d120b1a3ce359d8d3be8bda78524cd8f520c99883bfe33964cf
+
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends --yes jq \
+    && rm -rf /var/lib/apt/lists/*
--- a/.github/scripts/tests/test_pr_ci_workflow.py
+++ b/.github/scripts/tests/test_pr_ci_workflow.py
@@ -5,18 +5,21 @@ from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[3]
 WORKFLOW = ROOT / ".github" / "workflows" / "pr-ci.yml"
+SANDBOX_DOCKERFILE = ROOT / ".github" / "ci" / "Dockerfile"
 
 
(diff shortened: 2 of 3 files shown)


qiyuangong:
Self-hosted Runners have been moved to containers. PR not necessary. Closed.

PR #20, opened 2026-07-24
docs: restructure README run sections and clarify Quick Start

JasperChan929:
## Why the change

The root README mixed Docker-in-Docker usage into the FleetSpec section, front-loaded setup internals a first-time reader does not need, and gave no pointer to the valid taskset values. Review feedback on #10 asked for the same direction: keep details in scripts/README.md and the README to what users act on.

## Summary of the change

- Split Docker-in-Docker runs into their own section with a link to the dind-run.sh configuration and caveats; it applies to every launch mode, not just specs.
- Open the FleetSpec section with a one-sentence definition and a deep link to the full format (`scripts/README.md#fleetspec-json`).
- Trim the prerequisite list to packages that are not preinstalled (Docker, Python, git/curl/jq) and condense the setup paragraphs; the full list stays in scripts/README.md.
- Compress the launch/exit reporting paragraph to what the user sees on screen; RUN_ID, summary.txt, and exit-code details remain in scripts/README.md.
- Remove Harbor as a concept from the Quick Start flow and flag table; it stays in More details for readers going deeper.
- Link the taskset index (Tasks/README.md, now including TMax) and Tips & Caveats from More details.

## Other details

Documentation only; diff is README.md alone. Addresses the three review comments left on #10 (preinstalled packages, detail placement, run-reporting verbosity). All deep-link anchors verified against the current scripts/README.md and Tasks/README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,10 @@ evaluating Claude Code, OpenCode, and OpenClaw.
 - `git`, `curl`, `jq`, and `openssl`
 - Linux `util-linux` (`flock`, `setsid`, and `script`) and `procps` (`pkill`)
 
-These system commands must be on `PATH`; downloading an executable without
-adding its directory to `PATH` is not sufficient. `setup.sh` installs the
-tested Zellij and uv releases, Node.js, and Pi to user-writable locations and
-records their paths for every runner. The scripts do not require tmux. Harbor
-task containers still install and run the selected benchmark agent, including
-Claude Code.
+Install these with your system package manager, which puts them on `PATH`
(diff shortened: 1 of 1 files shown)


JasperChan929:
The three P2 findings on the prerequisites section are addressed in 38e71ef: the follow-up sentence no longer claims `setup.sh` installs the checked-only system tools (openssl, util-linux, procps) — it now says setup checks them and reports anything missing. The bullet list stays trimmed per the review feedback on #10.

qiyuangong (approved):
LGTM

PR #22, opened 2026-07-24
refactor(harbor): extract shared Pi subprocess runtime

AmaoLearning:
## 1. Why the change?

Harbor Analyzer currently owns Pi provider configuration, subprocess lifecycle management, JSON event parsing, final-output extraction, and provenance generation in a single component-specific module.

The staged Harbor Fixer work requires the same process and protocol mechanics. Duplicating this implementation would increase the initial Fixer PR substantially and allow runtime behavior—such as provider configuration, timeout handling, and event validation—to drift between Harbor components.

This PR extracts the component-independent mechanics into a small shared helper package while preserving the existing Analyzer behavior and interfaces.

## 2. Summary of the change

- Add `harbor_pi_runtime`, a small shared helper package responsible for:
  - isolated Pi subprocess startup and process-group timeout handling;
  - provider URL, environment, and `models.json` configuration;
  - raw or compact JSON event persistence;
  - Pi session and turn lifecycle validation;
  - provider-error and truncation detection;
  - final JSON extraction and runtime provenance.
- Convert `harbor_analyzer/pi.py` into a thin Analyzer-specific adapter.
- Keep Analyzer-specific behavior in the adapter, including:
  - read-only tools and path-gated extension configuration;
  - allowed-path and access-audit artifacts;
  - Analyzer prompts, output paths, return types, and error codes;
  - raw event retention and bearer authentication configuration.
- Add focused tests for shared runtime behavior and Analyzer compatibility.
- Update the Harbor structure documentation to describe the new package boundary.

The shared package intentionally does not contain Harbor task semantics, Analyzer/Fixer schemas, component prompts, or retry policy.

This PR does not migrate or otherwise modify Harbor Fixer. Fixer integration will be submitted in a subsequent staged PR.

## 3. Other details

Compatibility was checked against the existing Analyzer contracts:

- `dispatch_to_child()` and `DispatchResult` remain compatible.
- Analyzer artifact paths, raw event retention, tool restrictions, error codes, and provenance fields are preserved.
- Defensive handling remains at actual process and protocol boundaries, including process-group termination, malformed JSONL detection, lifecycle validation, provider failures, and truncated output.

Validation performed:

- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test_*.py'`
  - PR branch: 34 tests passed.
- A temporary, uncommitted Fixer adapter was used to validate the shared runtime against the latest `feat/fixer`.
  - Integrated Harbor suite: 54 tests passed.
- `git diff --check` passed.

The temporary Fixer adapter was discarded and is not part of this PR.

### DIFF
--- a/Agents/utils/common/Harbor/scripts/harbor_pi_runtime/process.py
+++ b/Agents/utils/common/Harbor/scripts/harbor_pi_runtime/process.py
@@ -1,0 +1,760 @@
+"""Launch isolated Pi subprocesses and validate their JSON event streams."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+import re
+import shutil
+import signal
+import subprocess
+import threading
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, BinaryIO
+from urllib.parse import urlparse
+
+ENV_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+MESSAGE_UPDATE_RE = re.compile(br'^\s*\{\s*"type"\s*:\s*"message_update"\s*[,}]')
+THINKING_LEVELS = {"off", "minimal", "low", "medium", "high", "xhigh", "max", "ultra"}
+PERSISTED_EVENT_TYPES = {
+    "agent_end",
+    "agent_start",
+    "auto_retry_end",
+    "auto_retry_start",
+    "message_end",
+    "result",
+    "session",
+    "turn_end",
+    "turn_start",
+}
+@dataclass
+class PiProcessResult:
+    output_json: dict[str, Any] | None
+    output_text: str
+    provenance: dict[str, Any]
+    block_reason: str | None
+    stderr_tail: str
+
+
+class _StreamingEventState:
+    def __init__(self) -> None:
+        self.hasher = hashlib.sha256()
+        self.event_count = 0
+        self.invalid_lines = 0
+        self.message_updates_dropped = 0
+        self.observed_bytes = 0
+        self.stored_bytes = 0
+        self.stream_error: OSError | None = None
+
+
+def _consume_compact_stream(
+    stream: BinaryIO,
+    output: BinaryIO,
+    state: _StreamingEventState,
+) -> None:
+    try:
+        for raw_line in iter(stream.readline, b""):
+            if not raw_line.strip():
+                continue
+            state.observed_bytes += len(raw_line)
+            if MESSAGE_UPDATE_RE.match(raw_line):
+                state.event_count += 1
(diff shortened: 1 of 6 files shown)


Agents/utils/common/Harbor/scripts/harbor_pi_runtime/process.py
```diff
@@ -0,0 +1,648 @@
...
+        return PiProcessResult(
+            None,
+            "",
+            record,
+            f"pi_event_compaction_error:{exc}",
+            stderr[-4000:],
+        )
+    finally:
+        if event_retention == "compact":
+            raw_events_path.unlink(missing_ok=True)
```
qiyuangong: **P2: Preserve raw events when compaction fails**

`compact_jsonl_event_stream()` can raise `OSError` after the child completes, but this `finally` block deletes the hidden raw stream regardless. I reproduced this by forcing a compaction failure: the helper returned `pi_event_compaction_error`, yet neither `events_path` nor `.events.jsonl.raw` remained, leaving the recorded provenance with no inspectable event artifact. Only unlink the raw stream after successful compaction; on failure, retain it and record its path or promote it to the public event path.

Agents/utils/common/Harbor/scripts/harbor_pi_runtime/process.py
```diff
@@ -0,0 +1,648 @@
...
+        "launch_mode": launch_mode,
+        "pi_binary": pi_bin,
+        "child_agent": agent_name,
+        "provider": provider,
+        "provider_api": "openai-completions",
+        "provider_base_url": normalized_url,
+        "api_key_env": api_key_env,
+        "events_path": str(events_path),
+        "independent_pi_process": True,
+        **(provenance or {}),
```
qiyuangong: **P2: Keep runtime-owned provenance authoritative**

Because caller-supplied `provenance` is expanded after the runtime facts, it can overwrite `provider`, `events_path`, `independent_pi_process`, and similar fields. A focused call with conflicting values completed successfully and returned those supplied values while executing with the real arguments, so audit records can silently disagree with reality. Merge caller metadata first and overwrite reserved fields from the actual arguments, or reject reserved-key collisions.

qiyuangong (commented):
Two reproducible P2 findings in the shared Pi runtime. The affected Harbor suite otherwise passes.

qiyuangong:
Need rebase before next review

qiyuangong:
@codex review

qiyuangong:
@codex review

qiyuangong:
@codex review

qiyuangong (approved):
LGTM

PR #27, opened 2026-07-25
Fix Ruff lint issues

qiyuangong:
## 1. Why the change?

Ruff 0.16.0 reports lint violations across the current Python codebase, and the repository does not yet pin a Ruff baseline.

## 2. Summary of the change

- Pin Ruff 0.16.0 and preserve the repository's intentional naming and import exceptions.
- Resolve the current Ruff findings with behavior-preserving Python cleanups.
- Preserve PinchBench's existing timezone-naive persisted timestamp format.
- Mark Python entry-point scripts with shebangs as executable.

## 3. Other details

Verification:

- `uvx ruff@0.16.0 check .`
- 96 GitHub script tests
- 37 Harbor tests
- 28 PinchBench tests
- 99 script tests (macOS-compatible tests locally; Bash 5 setup cases in Alpine Linux)
- Prerequisite and DinD shell tests
- Syntax validation for every tracked shell script

PR #33, opened 2026-07-25
Fix Ruff issues in RL utilities

qiyuangong:
## 1. Why the change?

The Ruff cleanup in #27 was too broad to review as one pull request. This PR isolates the RL rollout utilities.

## 2. Summary of the change

- Resolve Ruff findings in the remote Harbor service and worker utilities.
- Keep invalid request bodies mapped to HTTP 400 responses.
- Clean the RL request, startup, and worker tests.

## 3. Other details

This PR targets `main` independently from the other #27 replacements.

Verification:

- Targeted Ruff 0.16.0 check
- 8 RL utility tests


Agents/utils/rl/rollout_remote_harbor.py
```diff
@@ -494,7 +492,7 @@ def do_POST(self) -> None:  # noqa: N802
                 "duration_sec": round(time.monotonic() - started, 3),
             })
             self._send_json(HTTPStatus.OK, result)
-        except ValueError as exc:
+        except (TypeError, ValueError) as exc:
```
sgwhat: **[P1] Scope `TypeError` handling to request validation.** This `try` also covers `_enqueue_request()`, `_wait_result()`, tracing, and response serialization, so internal `TypeError`s are now misreported as client-side HTTP 400s. I reproduced that directly. There is also a concrete side-effect path: a truthy non-numeric `request_timeout` is converted only after `_enqueue_request()` has already queued the trial, then this handler returns 400; a client that corrects and retries can duplicate the work. Please validate/convert request fields before enqueueing and catch a dedicated bad-request exception (or narrowly catch around `_read_json()`). Add regressions asserting invalid timeouts do not enqueue and internal `TypeError`s still return 500.

sgwhat (commented):
Reviewed head `5783185`. One P1 behavior issue found; see the inline comment. The existing handler-test gap is valid, and the requested regressions below cover both the public contract and the internal-error boundary. The targeted 8 RL utility tests pass locally.

Agents/utils/rl/rollout_remote_harbor.py
qiyuangong: Fixed in d6b7112. Timeout parsing now happens before `_enqueue_request()`, so an invalid timeout cannot queue work. `TypeError` handling is limited to request parsing; internal `TypeError`s now reach the HTTP 500 boundary. Regressions cover both behaviors.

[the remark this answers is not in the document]
qiyuangong: Fixed in d6b7112. Added handler regressions for both non-object and malformed JSON bodies, asserting that each path returns HTTP 400 with the expected exception type.

Agents/utils/rl/rollout_remote_harbor.py
```diff
@@ -482,8 +500,18 @@ def do_POST(self) -> None:  # noqa: N802
         request_id = ""
         try:
             request = self._read_json()
-            request_id, result_path = _enqueue_request(request)
             wait_timeout = float(request.get("request_timeout") or request.get("timeout") or DEFAULT_TIMEOUT)
+        except (TypeError, ValueError) as exc:
```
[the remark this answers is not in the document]
qiyuangong: Fixed in 19dbdd2. Unexpected request-parsing exceptions now go through `_send_failure()` as structured HTTP 500 responses, while `TypeError`/`ValueError` validation failures remain HTTP 400. Added a `RecursionError` boundary regression.

Agents/utils/rl/rollout_remote_harbor.py
```diff
@@ -482,8 +500,27 @@ def do_POST(self) -> None:  # noqa: N802
...
+            self._send_failure(
+                HTTPStatus.BAD_REQUEST,
+                exc,
+                started=started,
+                request=request,
+                request_id=request_id,
```
sgwhat, line 510: [P2] Preserve the client request ID on pre-enqueue failures

Because timeout parsing now happens before `_enqueue_request()`, `request_id` is still empty here even when the parsed request supplied one. For example, `{"request_id": "trial-42", "request_timeout": "invalid"}` correctly returns 400 but records the error trace with `request_id: ""`, so operators cannot correlate that failure with the client request. Please copy the supplied request ID from `request` immediately after `_read_json()` (before parsing the timeout), or make `_send_failure()` fall back to it.

sgwhat (commented):
The original error-classification issue is fixed, and the new regression tests pass. I found one follow-up observability regression in the pre-enqueue validation path.

sgwhat (approved):
Approved. The error-classification fixes are verified and the regression tests pass. Please address the remaining request_id trace-correlation comment before merging.

PR #34, opened 2026-07-25
Fix Ruff issues in benchmark tasks

qiyuangong:
## 1. Why the change?

The Ruff cleanup in #27 was too broad to review as one pull request. This PR isolates PinchBench and ClawBio.

## 2. Summary of the change

- Resolve Ruff findings in PinchBench worker execution and summaries.
- Preserve PinchBench's timezone-naive persisted timestamp format.
- Resolve Ruff findings in the ClawBio runner and tests.

## 3. Other details

This PR targets `main` independently from the other #27 replacements.

Verification:

- Targeted Ruff 0.16.0 check
- 28 PinchBench tests
- 5 ClawBio tests (1 Linux-specific test skipped on macOS)

PR #28, opened 2026-07-25
Configure Ruff for core scripts

qiyuangong:
## 1. Why the change?

The Ruff cleanup in #27 was too broad to review as one pull request. This PR isolates the Ruff configuration and root script cleanup.

## 2. Summary of the change

- Pin Ruff 0.16.0 and preserve intentional naming, import, and test-fixture exceptions.
- Resolve Ruff findings in the root prompt script and its test suites.
- Mark the Python prompt entry point as executable.

## 3. Other details

This PR targets `main` independently from the other #27 replacements.

Verification:

- Targeted Ruff 0.16.0 check
- 4 root tests
- 99 script tests
- Prerequisite and DinD shell tests

PR #35, opened 2026-07-25
Add Ruff lint workflow

qiyuangong:
## 1. Why the change?

The repository is adopting Ruff 0.16.0 through cleanup PRs #28 through #34, but it does not yet enforce that lint baseline in GitHub Actions.

## 2. Summary of the change

- Add a dedicated GitHub-hosted Ruff workflow for pull requests and pushes to `main`.
- Run lint only with Ruff 0.16.0 and `.github/ruff.toml`.
- Pin `actions/checkout` and `astral-sh/ruff-action` to full commit SHAs.
- Add contract tests for triggers, permissions, runner selection, action pins, Ruff version, and lint-only arguments.

## 3. Other details

This draft depends on #28 through #34. Until those cleanup PRs merge, the full-repository Ruff job is expected to fail because `main` does not yet contain `.github/ruff.toml` or the complete lint cleanup.

Verification:

- `python3 -m unittest discover -s .github/scripts/tests -v` — 100 tests passed.
- `docker run --rm --volume "$PWD:/repo" --workdir /repo rhysd/actionlint:1.7.12 -color .github/workflows/ruff.yml`
- `uvx ruff@0.16.0 check --config .github/ruff.toml .` — passed against an assembled tree containing #28 through #34 and this PR.
- `git diff --check`

### DIFF
--- a/.github/scripts/tests/test_ruff_workflow.py
+++ b/.github/scripts/tests/test_ruff_workflow.py
@@ -34,10 +34,12 @@ class RuffWorkflowContractTest(unittest.TestCase):
             "278981a28ce3188b1e39527901f38254bf3aac89 # v4.1.0",
             self.workflow,
         )
-        self.assertEqual(
-            len(re.findall(r"uses: [^\s@]+@[0-9a-f]{40}", self.workflow)),
-            2,
+        action_refs = re.findall(r"uses:\s+([^\s#]+)", self.workflow)
+        pinned_action_refs = re.findall(
+            r"uses:\s+([^\s@]+@[0-9a-f]{40})(?=\s|$)",
+            self.workflow,
         )
+        self.assertEqual(action_refs, pinned_action_refs)
         self.assertIn("persist-credentials: false", self.workflow)
 
     def test_runs_only_pinned_ruff_lint_with_the_repository_config(self):
--- a/.github/scripts/tests/test_ruff_workflow.py
+++ b/.github/scripts/tests/test_ruff_workflow.py
@@ -1,0 +1,54 @@
+import re
+import unittest
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[3]
+WORKFLOW = ROOT / ".github" / "workflows" / "ruff.yml"
+
+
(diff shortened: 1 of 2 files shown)


.github/scripts/tests/test_ruff_workflow.py
```diff
@@ -0,0 +1,54 @@
...
+            " # v7.0.1",
+            self.workflow,
+        )
+        self.assertIn(
+            "uses: astral-sh/ruff-action@"
+            "278981a28ce3188b1e39527901f38254bf3aac89 # v4.1.0",
+            self.workflow,
+        )
+        self.assertEqual(
+            len(re.findall(r"uses: [^\s@]+@[0-9a-f]{40}", self.workflow)),
```
sgwhat: This assertion only proves that two pinned references exist; it does not prove that every `uses:` entry is pinned. If a later step adds `uses: owner/action@main`, the regex still finds the two existing SHA-pinned actions and this test remains green. Please compare the complete set of `uses:` entries against the full-SHA matches (or parse the workflow and validate every active step) so the pinning contract cannot silently regress.

qiyuangong: Fixed in 2114b90. The contract test now compares every `uses:` reference with the complete set of full-SHA matches. A mutation adding `owner/action@main` fails the test.

sgwhat: Thanks, the ordinary `uses: owner/action@main` mutation is now caught. There is still a valid-YAML bypass, though: both `"uses": owner/action@main` and `uses : owner/action@main` are accepted by `actionlint`, but neither regex sees those keys, so `action_refs == pinned_action_refs` remains true. I reproduced both against `2114b90`. Please parse the workflow and walk every step’s `uses` field (preferred), or support quoted/whitespace key variants and add these mutations as regression cases.

sgwhat (approved):
Approved. The workflow and dependency plan are sound. I will re-verify the remaining YAML `uses`-key pinning regression on the next commit before merge.

PR #39, opened 2026-07-27
fix(opencode): stop faking git for non-repository tasks

### LINKED ISSUE #36: OpenCode completed runs exit non-zero and trigger unnecessary Harbor retries
Migrated from [`sii-system/tasks#138`](https://github.com/sii-system/tasks/issues/138). GitHub cannot natively transfer an issue from a private repository to this public repository, so this is a sanitized, updated copy.

Related: [`sii-system/tasks#117`](https://github.com/sii-system/tasks/issues/117) (Harbor agent adapter reliability).

## Summary

OpenCode benchmark runs against `custom/glm-5.1-fp8` can complete the requested work and receive verifier reward `1`, while the OpenCode process exits non-zero. Harbor classifies the run as `NonZeroAgentExitCodeError` and retries work that already succeeded.

The issue remains present in `agent-fleet` at [`af390bd`](https://github.com/sii-system/agent-fleet/commit/af390bdf22775c9d8a29041b25d25a1f9a3558f0) (verified 2026-07-27).

## Current code state

The adapter still preserves the raw pipeline status after running the trace finalizer:

```bash
opencode ... run --format=json ... | tee /logs/agent/opencode.txt
opencode_rc=$?
python3 finalize_opencode_sessions.py ... || true
exit "$opencode_rc"
```

Relevant code:

- [`opik_opencode_harbor.py`](https://github.com/sii-system/agent-fleet/blob/af390bdf22775c9d8a29041b25d25a1f9a3558f0/Agents/Harbor-opencode/opik_opencode_harbor.py#L592-L604) propagates the raw OpenCode status.
- [`finalize_opencode_sessions.py`](https://github.com/sii-system/agent-fleet/blob/af390bdf22775c9d8a29041b25d25a1f9a3558f0/Agents/Harbor-opencode/finalize_opencode_sessions.py#L133-L218) finalizes Opik traces; it does not determine whether the agent session completed successfully. Its default `status = "completed"` is trace metadata, not a completion check.
- [`env.sh`](https://github.com/sii-system/agent-fleet/blob/af390bdf22775c9d8a29041b25d25a1f9a3558f0/Agents/utils/common/Harbor/env.sh#L220-L230) does not exclude `NonZeroAgentExitCodeError` from default retries.
- No `check_session_completed.py` or equivalent completion gate currently exists.

## Existing run evidence

The latest available OpenCode smoke artifacts on the runner host are from 2026-07-17:

- Four standard OpenCode smoke runs recorded verifier reward `1` together with `NonZeroAgentExitCodeError` (`exit 137`). Their job logs show Harbor retrying the failed trial.
- A separate `--pure` control also exited `137`, but received reward `0`.

This demonstrates why a blanket `exit 0` is unsafe: the adapter must distinguish completed work from genuine crashes, incomplete sessions, and failed runs.

### Freshness caveat

Those artifacts used cached OpenCode `1.17.14`, although the cache filename and configuration used the tag `latest`. The npm `latest` release is now [`1.18.5`](https://registry.npmjs.org/opencode-ai/latest), whose release notes do not explicitly identify a fix for this end-of-session behavior.

Before implementing normalization, refresh the OpenCode cache and rerun one smoke while recording `opencode --version`. Even if a newer OpenCode version no longer reproduces the provider-specific hang, the adapter currently has no safe handling for completed sessions that end with a dirty CLI status.

## Proposed fix

Keep the change inside `Agents/Harbor-opencode/` and add a conservative, tracing-independent session-completion check. Do not normalize every non-zero status.

```bash
opencode_rc=$?
python3 finalize_opencode_sessions.py ... || true

if [ "$opencode_rc" -ne 0 ] && python3 check_session_completed.py \
    --event-log /logs/agent/opencode.txt; then
  echo "[INFO] opencode rc=$opencode_rc but session completion was confirmed; normalizing to 0"
  opencode_rc=0
fi

exit "$opencode_rc"
```

The checker may parse the JSON event stream or OpenCode session storage, but it must:

- return success only when there is positive evidence that the main agent session reached a terminal, completed state;
- return non-zero for missing, malformed, incomplete, or ambiguous state;
- work when `TRACE_TO_OPIK=false` and avoid depending on the Opik plugin state file;
- log the original OpenCode return code and the evidence used for normalization without exposing prompts, credentials, or provider configuration.

If OpenCode does not expose a reliable terminal state for the observed exit-137 case, do not guess. Move the decision to the narrowest post-verifier layer where a valid reward is available, rather than treating any tool completion as full task completion.

## Acceptance criteria

- Non-zero OpenCode status plus confirmed completed session is normalized to `0`.
- Non-zero status plus incomplete, malformed, missing, or ambiguous session data preserves the original status.
- A normal zero exit remains unchanged.
- The behavior is identical with tracing enabled and disabled.
- Unit tests cover all paths above, including a genuine crash/timeout fixture.
- A refreshed real smoke against `custom/glm-5.1-fp8` records the actual OpenCode version and demonstrates one of:
  - clean CLI exit with no retry; or
  - dirty CLI exit normalized only after positive completion evidence, with no retry.
- The normalization log is monitored for several batches to detect provider-side regressions or false positives.

## Interim mitigation

Until the guarded fix lands, operators can suppress retries in `config.local.env`:

```bash
TB_RETRY_EXCLUDE_EXCEPTIONS=RewardFileNotFoundError,RewardFileEmptyError,VerifierOutputParseError,NonZeroAgentExitCodeError
```

This also suppresses retries for genuine OpenCode crashes, so it should remain an explicit temporary override rather than becoming the default.

JasperChan929:
## 1. Why the change?

Harbor's OpenCode adapter forced `OPENCODE_FAKE_VCS=git` for every task. On
non-repository task images, OpenCode consequently treated `/` as a Git
worktree and snapshotted the entire container with `git add --all`. Under the
task's 1 GiB memory limit, the second snapshot pass was OOM-killed after the
tool had already completed, so successful tasks exited 137 and triggered
unnecessary Harbor retries.

Fixes #36.

## 2. Summary of the change

- Stop injecting OpenCode's testing-only fake-VCS flag.
- Add a regression assertion that the adapter does not restore the flag.

OpenCode still detects real Git repositories normally, so repository-scoped
snapshot and undo behavior remains available.

## 3. Other details

Validation:

- `python3 -m unittest discover -s Agents/Harbor-opencode/tests -p 'test_*.py'`
  (9 tests passed)
- OpenCode 1.18.5 baseline, 1 GiB, zero retries: 4m47s, Docker `container oom`,
  exit 137, reward 1, one `NonZeroAgentExitCodeError`
- Same run with this change: 34s, reward 1, zero exceptions, two
  `step_finish` events, no snapshot hash, and no Docker OOM event

A trace-enabled live run was blocked before task startup by a TLS EOF from the
remote Opik health endpoint. The trace-enabled adapter path remains covered by
the unit suite.

### DIFF
--- a/Agents/Harbor-opencode/opik_opencode_harbor.py
+++ b/Agents/Harbor-opencode/opik_opencode_harbor.py
@@ -536,7 +536,6 @@ class OpikOpenCodeHarbor(OpenCode):
                 if key in env:
                     env[key] = _rewrite_container_opik_url(env[key])
 
-        env["OPENCODE_FAKE_VCS"] = "git"
         if trace_enabled:
             # Harbor only downloads EnvironmentPaths.agent_dir after timeout.
             # Keep the hook runtime backup there so the outer worker can replay
--- a/Agents/Harbor-opencode/tests/test_trace_disabled.py
+++ b/Agents/Harbor-opencode/tests/test_trace_disabled.py
@@ -179,6 +179,10 @@ class OpenCodeTraceDisabledTests(unittest.TestCase):
             "OC_OPIK_LOGS_DIR",
             agent.agent_commands[-1].get("env", {}),
         )
+        self.assertNotIn(
+            "OPENCODE_FAKE_VCS",
+            agent.agent_commands[-1].get("env", {}),
+        )
 
     def test_run_trace_on_keeps_plugin_registration_and_finalizer(self) -> None:
         agent = self.make_agent("true")


qiyuangong (approved):
LGTM

PR #43, opened 2026-07-27
Fail fast on incompatible ClawBio security settings

### LINKED ISSUE #41: Fail fast when ClawBio security settings block benchmark execution
## Why

A real ClawBio canary can pass task selection and fleet startup but fail only after task execution begins when it inherits the general OpenClaw safety defaults:

- `EXEC_SECURITY=deny`
- `WORKSPACE_ONLY=true`

ClawBio needs to read its plugin skill outside the instance workspace, execute shell/Python/R commands, and write benchmark artifacts. With the restrictive defaults, the run reports errors such as `Path escapes sandbox root` and `exec denied`.

The ClawBio README currently documents the required benchmark settings (`EXEC_SECURITY=full`, `EXEC_ASK=off`, `WORKSPACE_ONLY=false`, and `DOCKER_COMPOSE_READ_ONLY=false`), but the unified launcher does not validate them before performing setup and starting the fleet.

## Proposed behavior

- Keep the general OpenClaw security defaults unchanged.
- Add an explicit ClawBio benchmark mode, or validate the required settings at launcher startup.
- Fail with a clear explanation before cache/image/fleet mutation when the selected settings cannot execute ClawBio.
- Require explicit opt-in before relaxing execution or workspace restrictions.

## Acceptance criteria

- The unified ClawBio launcher detects incompatible security settings before starting benchmark work.
- The error identifies the relevant settings and points to the documented benchmark configuration.
- Normal OpenClaw defaults are not weakened globally.
- Tests cover both the restrictive failure path and the explicit benchmark path.

JasperChan929:
## 1. Why the change?

The ClawBio unified launcher could inherit incompatible OpenClaw execution settings, prepare caches and images, regenerate and restart the fleet, and only then fail during task execution with errors such as `exec denied` or `path escapes sandbox`. Users received the actionable configuration error after expensive work and fleet side effects had already occurred.

Closes #41.

## 2. Summary of the change

- Add a committed, public-safe `Tasks/clawBio/config/benchmark.env` with the four settings ClawBio requires: `SANDBOX_MODE=off`, `EXEC_SECURITY=full`, `EXEC_ASK=off`, and `WORKSPACE_ONLY=false`.
- Load that dedicated profile after shared and fleet defaults, while keeping runtime environment variables as the highest-precedence overrides.
- Warn users that the profile is permissive and remains active in the generated fleet until it is stopped or regenerated.
- Reject conflicting overrides before run-root creation, image build, cache prewarm, setup, or fleet restart.
- Keep task-selection-only validation side-effect free and independent of runtime security setup.
- Use the simple ClawBio and FleetSpec launch commands in documentation and source the same profile for manual setup.

## 3. Other details

The general OpenClaw defaults remain restrictive and unchanged. ClawBio setup writes the benchmark-only profile into its run-specific generated configs and leaves that fleet running; the launcher and documentation explicitly tell users to stop or regenerate it before other use.

A real `rnaseq-de-demo` canary completed successfully with 11 artifacts while Docker reported `ReadonlyRootfs=true`. Therefore `DOCKER_COMPOSE_READ_ONLY=false` is not required: the container root filesystem remains read-only by default, while ClawBio writes through its dedicated state/workspace mounts and `/tmp` tmpfs.

Validation:

- `python3 -m unittest discover -s Tasks/clawBio/tests` — 10 passed
- `python3 -m unittest discover -s scripts/tests` — 113 passed
- `uvx --offline ruff==0.16.0 check --config .github/ruff.toml`
- Bash syntax and `git diff --check`
- Direct profile-loading smoke check: warning emitted, no run root created before the later Opik preflight
- Real ClawBio canary: `rnaseq-de-demo`, 1/1 success, 11 artifacts, read-only root filesystem

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -172,7 +172,8 @@ Examples:
   --task fix-git,break-filter-js-from-html --workers 2
 ./scripts/run_fleet.sh --taskset ./my-taskset --agent opencode --workers 2
 ./scripts/run_fleet.sh --taskset pinchbench --task task_sanity --workers 1
-./scripts/run_fleet.sh --taskset clawbio --task rnaseq-de-demo --workers 1
+EXEC_SECURITY=full EXEC_ASK=off WORKSPACE_ONLY=false \
+  ./scripts/run_fleet.sh --taskset clawbio --task rnaseq-de-demo --workers 1
 ./scripts/run_fleet.sh --taskset terminal-bench/terminal-bench-2-1 \
   --agent claude-code --workers 10 --output fleet-spec.json --dry-run
 ```
@@ -414,11 +415,13 @@ workers to `--instances`, and validates selected IDs against the pinned
 `task_*.md` checkout before image or result creation; the OpenClaw fleet must
 already be configured and running. The `clawbio` taskset calls the existing
 ClawBio unified launcher, maps workers to `COUNT`, and validates selected IDs
-against its task config before setup or fleet restart. Those runners own setup,
-execution, outputs, and failures. If `--agent` conflicts with an OpenClaw
-taskset, the router prints the requested and actual agents, ignores the
-conflicting value, and continues with OpenClaw. OpenClaw runners remain in the
-foreground; `--detach` is ignored with a warning.
+against its task config before setup or fleet restart. The launcher also
+requires the explicit ClawBio benchmark security settings documented in
+`Tasks/clawBio/README.md`; FleetSpec does not store environment variables.
+Those runners own setup, execution, outputs, and failures. If `--agent`
+conflicts with an OpenClaw taskset, the router prints the requested and actual
+agents, ignores the conflicting value, and continues with OpenClaw. OpenClaw
+runners remain in the foreground; `--detach` is ignored with a warning.
 
 ### Current limitations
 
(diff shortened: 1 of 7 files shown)


JasperChan929:
@codex review

JasperChan929:
@codex review

JasperChan929:
@codex review

scripts/README.md
```diff
@@ -172,7 +172,8 @@ Examples:
   --task fix-git,break-filter-js-from-html --workers 2
 ./scripts/run_fleet.sh --taskset ./my-taskset --agent opencode --workers 2
 ./scripts/run_fleet.sh --taskset pinchbench --task task_sanity --workers 1
-./scripts/run_fleet.sh --taskset clawbio --task rnaseq-de-demo --workers 1
+SANDBOX_MODE=off EXEC_SECURITY=full EXEC_ASK=off WORKSPACE_ONLY=false \
```
qiyuangong: If `SANDBOX_MODE=off EXEC_SECURITY=full EXEC_ASK=off WORKSPACE_ONLY=false` are necessary for clawbio. Maybe we should add it to clawbio config, and warn user about that setting. Rather than change launch command

JasperChan929:
@codex review

qiyuangong (approved):
LGTM

PR #58, opened 2026-07-29
Fix Harbor benchmark PID identity mismatch

### LINKED ISSUE #52: Harbor monitor misclassifies native registry runs as abnormal_exit due to mismatched PID start time
## Summary

Native Harbor registry runs can be misclassified as `abnormal_exit` while the benchmark process and task containers are still running. The monitor then exits with `stop_action_required`, and the analyzer supervisor shuts down before receiving the final failed-task handover.

This is an integration-contract bug between the PID identity producer in `Agents/utils/common/Harbor/harboropik.sh` and the consumer in `scripts/harbor_monitor/runner.py`; standalone monitor tests with a correctly constructed PID file do not expose it.

## Affected path

The faulty writer runs when all of the following are true:

- `ROLLOUT != 1`
- the dataset is a registry dataset, for example `terminalbench21`
- `HARBOR_QUEUE_WORKER != 1`

One reproducing entrypoint is:

```bash
./scripts/run_fleet.sh \
  --taskset terminalbench21 \
  --agent claude-code \
  --workers 10 \
  --detach
```

## Actual behavior

In an observed run:

- Harbor remained alive and ultimately completed all 10 trials.
- At 20:50:23, the monitor reported:
  - `benchmark_status=blocked`
  - `status_reason=abnormal_exit`
  - `finished=1`
  - `unfinished=9`
  - `running_workers=0`
  - `monitor_follow_decision=stop_action_required`
- The Harbor run later completed normally at 21:48:14 with `completed=10`, `errors=0`.
- The analyzer only published its initial empty report with `subagent_skipped=true` and `task_count=0`; it never analyzed the three reward-zero tasks.

## Root cause

`harboropik.sh` writes the benchmark PID identity with this expression:

```bash
printf '%s\t%s\n' \
  "${BASHPID:-$$}" \
  "$(awk '{print $22}' "/proc/${BASHPID:-$$}/stat")" \
  > "$HARBOR_BENCHMARK_PID_FILE"
```

The command substitution runs in a Bash subshell. Inside `$(...)`, `BASHPID` refers to that subshell rather than the outer Harbor wrapper. The file can therefore contain:

- PID: outer Harbor wrapper Bash
- start time: command-substitution child

`process_is_alive()` correctly compares `/proc/<recorded-pid>/stat` field 22 with the recorded start time. Because the producer wrote a mismatched pair, it returns false even though the outer process is alive.

A minimal reproduction of the original expression produced different values (`match=no`). Depending on scheduling, parent and child may occasionally start in the same clock tick, which can mask the bug.

## Expected behavior

The PID and start time written to `HARBOR_BENCHMARK_PID_FILE` must describe the same process. The monitor should continue following the run until Harbor produces a terminal result or a real exit condition occurs, and the analyzer should receive the final failed-task handover.

## Suggested fix

Capture the outer PID before entering command substitution and reuse that stable value:

```bash
harbor_benchmark_pid="${BASHPID:-$$}"
printf '%s\t%s\n' \
  "$harbor_benchmark_pid" \
  "$(awk '{print $22}' "/proc/$harbor_benchmark_pid/stat")" \
  > "$HARBOR_BENCHMARK_PID_FILE"
```

## Regression coverage

Extend the native-registry `harboropik.sh` integration test so the fake Harbor executable validates, while the wrapper is still alive, that:

```text
recorded_start_time == /proc/<recorded_pid>/stat field 22
```

This coverage belongs at the producer/consumer integration boundary; monitor-only fixtures with valid PID identities will not catch the faulty writer.

0407nancy:
## Summary

- capture the outer Harbor wrapper PID before entering command substitution
- write the PID and `/proc/<pid>/stat` start time from the same process
- add native-registry integration coverage for the PID identity contract

## Root cause

`BASHPID` was evaluated again inside `$(...)`, where it referred to the command-substitution child rather than the outer Harbor wrapper. The PID file could therefore contain the wrapper PID paired with the child process start time, causing the monitor to treat a live benchmark as an abnormal exit.

Fixes https://github.com/sii-system/agent-fleet/issues/52

## Testing

- `bash -n Agents/utils/common/Harbor/harboropik.sh Agents/utils/common/Harbor/tests/test_harboropik_extra_compose.sh`
- `bash Agents/utils/common/Harbor/tests/test_harboropik_extra_compose.sh`
- `git diff --check`

### DIFF
--- a/Agents/utils/common/Harbor/harboropik.sh
+++ b/Agents/utils/common/Harbor/harboropik.sh
@@ -65,7 +65,11 @@ if harbor_is_native_registry_main; then
   : > "$HARBOR_JOB_DIR_FILE"
   rm -f "$HARBOR_BENCHMARK_EXIT_FILE"
   # BASHPID needs bash >= 4; at this top-level scope $$ is the same pid.
-  printf '%s\t%s\n' "${BASHPID:-$$}" "$(awk '{print $22}' "/proc/${BASHPID:-$$}/stat")" > "$HARBOR_BENCHMARK_PID_FILE"
+  harbor_benchmark_pid="${BASHPID:-$$}"
+  printf '%s\t%s\n' \
+    "$harbor_benchmark_pid" \
+    "$(awk '{print $22}' "/proc/$harbor_benchmark_pid/stat")" \
+    > "$HARBOR_BENCHMARK_PID_FILE"
   record_harbor_benchmark_exit() {
     local rc="$?"
     # The exit file is the completion contract with the registry monitor;
--- a/Agents/utils/common/Harbor/tests/test_harboropik_extra_compose.sh
+++ b/Agents/utils/common/Harbor/tests/test_harboropik_extra_compose.sh
@@ -75,6 +75,16 @@ inherited = {name: os.environ[name] for name in connection_fields if name in os.
 Path(f"{capture}.opik-environment").write_text(json.dumps(inherited, sort_keys=True))
 
 if os.environ.get("HARBOR_CAPTURE_RESULT") == "1":
+    pid_file = Path(os.environ["HARBOR_BENCHMARK_PID_FILE"])
+    if pid_file.is_file():
+        recorded_pid, recorded_start_time = pid_file.read_text().split()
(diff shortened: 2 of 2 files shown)


qiyuangong (approved):
LGTM

PR #59, opened 2026-07-29
Stream Harbor analyzer Pi events

### LINKED ISSUE #53: Harbor analyzer event log grows quadratically on cumulative Pi message_update events
## Summary

The Harbor analyzer stores raw `pi --mode json --print` stdout directly as `analyzer-subagent-events/<analysis_id>.jsonl`. Pi's `message_update` events include the cumulative assistant message state, so each streaming update repeats an increasingly large payload. For long analyzer turns this makes the JSONL grow approximately quadratically relative to the actual model output.

The analyzer then reads the entire file into memory to hash and parse it, amplifying the disk/IO issue into a memory-risk issue.

## Observed reproduction

Run: `2026-07-28-2209-harbor-tui`

Analyzer stack:

- Pi version: `0.81.1`
- dispatch mode: `per_task_pi_subagent`
- provider API: `openai-completions`
- affected task analysis completed successfully; this was not a provider/network failure

For one `dna-insert` analysis:

- raw event JSONL: `841,578,326` bytes (`du`: 803 MiB)
- total JSONL events: `16,414`
- `message_update` events: `16,214`
- bytes attributable to `message_update`: `838,630,486` (>99%)
- final analyzer JSON: `52,153` bytes
- benchmark report: `42,966` bytes

A sampled sequence demonstrates the cumulative payload:

| Event | JSONL line bytes | Full message content chars |
| --- | ---: | ---: |
| first `message_update` | 1,043 | 3 |
| update around 25% | 29,354 | 13,794 |
| update around 50% | 78,280 | 37,655 |
| late update | 161,589 | 55,243 |

The run produced a correct final classification, but retained roughly 842 MB of raw streaming snapshots to produce a ~52 KB final artifact.

## Code path

`Agents/utils/common/Harbor/scripts/harbor_analyzer/pi.py`:

1. `dispatch_to_child()` launches Pi and redirects stdout directly to `events_path`.
2. Every cumulative `message_update` snapshot is retained without filtering, compression, rotation, or a size bound.
3. After Pi exits, `events_path.read_text()` loads the entire JSONL into memory.
4. Hashing encodes that full string again, and `_parse_jsonl(raw)` calls `splitlines()` and parses all events, creating additional transient copies.

## Impact

- Hundreds of MiB of disk use for a single analyzer task.
- Heavy filesystem IO and slower completion/status inspection.
- High transient memory use after the child exits.
- Larger analyzer outputs can plausibly exhaust disk or memory even though the final result is small.
- Artifact retention becomes impractical for benchmark batches.

## Expected behavior

Analyzer evidence size should grow approximately linearly with meaningful output and should have a defined upper bound. Final-output extraction and provenance accounting should not require loading the complete raw stream into memory.

## Suggested fix

- Consume Pi stdout as a stream instead of redirecting it unchanged.
- For `message_update`, persist only the actual delta from `assistantMessageEvent`, or omit intermediate updates while retaining `message_start`, `message_end`, `turn_end`, lifecycle, retry, and tool-execution events.
- Compute the event hash incrementally.
- Extract lifecycle counts, provider errors, and the final assistant message incrementally rather than using `read_text()` + `splitlines()`.
- Add a configurable byte cap and explicit provenance such as `events_truncated`, `message_updates_dropped`, and original/stored byte counts.
- Optionally compress retained raw evidence.

## Acceptance criteria

- A synthetic stream of cumulative `message_update` snapshots produces stored evidence proportional to deltas/final content, not the sum of all cumulative snapshots.
- Final JSON extraction and lifecycle/provider-error validation remain correct.
- Analyzer does not read the entire event file into memory after completion.
- Tests cover size bounds, truncation metadata, malformed JSONL, timeouts, and successful final-report extraction.

0407nancy:
## Summary

- consume Pi JSONL stdout as a stream instead of redirecting and rereading the complete event file
- drop cumulative `message_update` snapshots while retaining lifecycle, retry, tool, and final-message events
- compute event hashes, byte counts, lifecycle state, provider errors, and final output incrementally
- preserve the existing timeout and final-report validation behavior

## Root cause

Pi emits cumulative assistant state in each `message_update`. Persisting every snapshot makes analyzer event artifacts grow approximately quadratically, and rereading the complete JSONL after Pi exits adds substantial transient memory usage.

The retained evidence now grows with meaningful events and final output. No fixed artifact cap is added because cumulative updates are discarded at the source and introducing truncation would add separate output semantics.

Fixes https://github.com/sii-system/agent-fleet/issues/53

## Testing

- `python3 -m py_compile Agents/utils/common/Harbor/scripts/harbor_analyzer/pi.py Agents/utils/common/Harbor/tests/test_harbor_analyzer_runtime.py`
- `python3 -m unittest Agents.utils.common.Harbor.tests.test_harbor_analyzer_runtime` (20 tests)
- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test_*.py'` (55 tests)
- `git diff --check`

### DIFF
--- a/Agents/utils/common/Harbor/scripts/harbor_analyzer/pi.py
+++ b/Agents/utils/common/Harbor/scripts/harbor_analyzer/pi.py
@@ -9,14 +9,16 @@ import re
 import shutil
 import signal
 import subprocess
-from dataclasses import dataclass
+import threading
+from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any
+from typing import Any, BinaryIO
 from urllib.parse import urlparse
 
 from .io import canonical_json, write_text_atomic
 
 ENV_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+MESSAGE_UPDATE_RE = re.compile(br'^\s*\{\s*"type"\s*:\s*"message_update"\s*[,}]')
 PI_EXTENSION_PATH = Path(__file__).resolve().parent / "pi_extensions" / "analyzer_path_gate.ts"
 
 
@@ -129,24 +131,6 @@ def _models_config(
     }
 
 
-def _parse_jsonl(raw: str) -> tuple[list[dict[str, Any]], int]:
-    events: list[dict[str, Any]] = []
-    invalid_lines = 0
-    for line in raw.splitlines():
-        if not line.strip():
-            continue
-        try:
-            event = json.loads(line)
-        except json.JSONDecodeError:
-            invalid_lines += 1
-            continue
-        if isinstance(event, dict):
-            events.append(event)
-        else:
-            invalid_lines += 1
-    return events, invalid_lines
-
-
 def _message_text(message: Any) -> str:
     if not isinstance(message, dict) or message.get("role") != "assistant":
         return ""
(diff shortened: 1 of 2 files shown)


qiyuangong (approved):
LGTM

PR #60, opened 2026-07-29
Run DinD setup as the host-aligned user

### LINKED ISSUE #57: DinD bootstrap tries to clone over an existing bind-mounted repository after Git dubious ownership failure
## Summary

A fresh DinD bootstrap can misclassify the already bind-mounted Agent Fleet checkout as a missing Git repository.

`scripts/dind-run.sh` runs `scripts/setup.sh` as root, while the mounted checkout is owned by the non-root host/DinD user. Git rejects the checkout with its `safe.directory` / dubious ownership check.

`setup.sh` suppresses the `git rev-parse` error and treats every failure as meaning that the repository does not exist. It then attempts to clone into the existing non-empty bind mount and exits with:

```text
[INFO] Cloning repo to /home/ziyi/agent-fleet-backup...
fatal: destination path '/home/ziyi/agent-fleet-backup' already exists and is not an empty directory.
```

## Affected path

The failure occurs when all of the following are true:

- Agent Fleet is launched through `scripts/dind-run.sh`.
- `DIND_BOOTSTRAP=always`, or `DIND_BOOTSTRAP=missing` decides setup is required.
- The repository is bind-mounted from a non-root host user.
- `dind-run.sh` executes `setup.sh` as root.
- The persistent DinD home does not already contain a matching Git `safe.directory` entry.

One reproducing entrypoint is:

```bash
./scripts/dind-run.sh \
  --taskset terminalbench21 \
  --agent claude-code \
  --workers 1
```

The issue is most visible with a fresh `DIND_HOME_VOLUME`.

## Actual behavior

The repository exists and contains a valid `.git` directory, but this probe fails when executed by root without a `safe.directory` entry:

```bash
git -C /home/ziyi/agent-fleet-backup rev-parse --git-dir
```

Git returns:

```text
fatal: detected dubious ownership in repository at '/home/ziyi/agent-fleet-backup'

To add an exception for this directory, call:

git config --global --add safe.directory /home/ziyi/agent-fleet-backup
```

`setup.sh` currently uses:

```bash
if git -C "$REPO_DIR" rev-parse --git-dir >/dev/null 2>&1; then
  ok "Repo already exists: $REPO_DIR (skip clone)"
else
  info "Cloning repo to $REPO_DIR..."
  git clone --recurse-submodules "$REPO_URL" "$REPO_DIR"
fi
```

Because stderr is discarded and all nonzero results enter the clone branch, the Git trust failure is incorrectly treated as a missing repository.

Adding the exact mounted path as a safe directory makes the same checkout pass the probe and allows setup to continue:

```bash
git config --global --add safe.directory \
  /home/ziyi/agent-fleet-backup
```

## Root cause

There is an integration mismatch between the DinD launcher and the setup repository detection:

1. `dind-run.sh` bind-mounts the host checkout.
2. It aligns `DIND_USER` with the host UID, so the checkout remains owned by the non-root user.
3. It then invokes setup as root:

```bash
docker exec "$DIND_NAME" env ... ./scripts/setup.sh
```

4. No Git trust entry is installed for the bind-mounted checkout.
5. `setup.sh` assumes every failed `git rev-parse` means the checkout is absent.

The local workaround persists in `DIND_HOME_VOLUME`, which can mask the bug on later recreations.

## Expected behavior

A valid checkout deliberately bind-mounted by `dind-run.sh` should be recognized as the existing repository during bootstrap.

`setup.sh` must not attempt to clone over an existing non-empty path merely because Git rejected the checkout for ownership reasons.

If the repository cannot be trusted or inspected, setup should either configure the exact trusted bind mount or stop with the original actionable Git error.

## Suggested fix

Before running `setup.sh` as root, register only the exact bind-mounted repository path as safe in the Git configuration used by that setup process:

```bash
git config --global --add safe.directory "$REPO_ROOT"
```

The implementation should avoid duplicate entries on repeated runs and must not use the broad unsafe form:

```bash
safe.directory=*
```

A command-scoped Git configuration passed only to `setup.sh` would also avoid permanently changing global Git trust settings.

Additionally, harden the clone decision in `setup.sh`:

- If `$REPO_DIR/.git` exists but `git rev-parse` fails, report the Git diagnostic and exit.
- Do not fall through to `git clone` when `$REPO_DIR` is already non-empty.

## Regression coverage

Extend the DinD tests so that:

1. The mounted repository is owned by a non-root UID.
2. Bootstrap executes `setup.sh` as root.
3. No pre-existing `safe.directory` entry is present.
4. The existing checkout is recognized and no clone is attempted.
5. Repeated bootstrap runs do not add duplicate trust entries.
6. An unrelated non-empty non-repository directory still fails clearly instead of being overwritten.

Also add a setup-level test where `git rev-parse` returns a dubious-ownership-style failure and verify that `git clone` is never called.

JasperChan929:
## 1. Why the change?

DinD aligns its benchmark user with the host UID/GID and bind-mounts the
checkout, but setup was the one launcher step still executed as container
root. Git then rejected the host-owned checkout as dubious, while `setup.sh`
treated every failed repository probe as a missing repository and attempted
to clone over the populated mount. Root setup could also leave artifacts that
the benchmark, Monitor, and Analyzer user could not update.

Fixes #57.

## 2. Summary of the change

- Prepare the DinD home volume once as `DIND_USER:DIND_USER`, then run setup
  through the same `--user "$DIND_USER"` path as `run_fleet.sh`.
- Remove the root setup helper and the post-setup recursive ownership repair.
- Make repository detection fail closed when `.git` exists, the destination is
  inaccessible, or the destination is non-empty; retain the original Git
  diagnostic instead of falling through to `git clone`.
- Preserve cloning for missing or empty destinations and keep recursive
  submodule sync/update behavior.
- Add regression coverage for non-root setup, ownership preparation, existing
  checkout reuse, dubious-ownership failure, non-repository populated
  destinations, and the valid empty-destination clone path.

## 3. Other details

Validation:

- `python3 -m unittest scripts/tests/test_setup.py` (8 tests)
- Bash syntax checks and `git diff --check`
- Fresh DinD container and volumes with fake credentials, completing setup,
  recursive submodule initialization, Docker permission validation, and
  `run_fleet.sh --dry-run`
- Verified setup outputs in the home volume and checkout were owned by the
  host-aligned `DIND_USER` UID/GID and remained writable

The validation checkout is a linked worktree, so the successful fresh
bootstrap mounted an isolated `/tmp` copy of its Git metadata; the original
worktree was not modified. Test containers, volumes, fake configuration, and
temporary metadata were removed afterward.

A full benchmark using real credentials was intentionally not run because the
separate DinD credential-argv hardening is outside this PR. The stale-home
bootstrap probe and credential transport defects are also intentionally left
for follow-up changes. No real secret was used or recorded.

### DIFF
--- a/scripts/dind-run.sh
+++ b/scripts/dind-run.sh
@@ -355,7 +355,7 @@ if [[ "$DIND_USER" != "root" ]]; then
   if [[ "$current_uid" != "$DIND_USER_UID" || "$current_gid" != "$DIND_USER_GID" ]]; then
     docker exec "$DIND_NAME" usermod -o -u "$DIND_USER_UID" -g "$DIND_USER_GID" "$DIND_USER"
   fi
-  docker exec "$DIND_NAME" chown -R "$DIND_USER:$DIND_USER" "$DIND_HOME_DIR"
+  docker exec "$DIND_NAME" chown -R "$DIND_USER_UID:$DIND_USER_GID" "$DIND_HOME_DIR"
 fi
 
 declare -a exec_base=(exec --user "$DIND_USER")
--- a/scripts/dind-run.sh
+++ b/scripts/dind-run.sh
@@ -355,7 +355,7 @@ if [[ "$DIND_USER" != "root" ]]; then
   if [[ "$current_uid" != "$DIND_USER_UID" || "$current_gid" != "$DIND_USER_GID" ]]; then
     docker exec "$DIND_NAME" usermod -o -u "$DIND_USER_UID" -g "$DIND_USER_GID" "$DIND_USER"
   fi
-  docker exec "$DIND_NAME" chown -R "$DIND_USER" "$DIND_HOME_DIR"
+  docker exec "$DIND_NAME" chown -R "$DIND_USER:$DIND_USER" "$DIND_HOME_DIR"
 fi
 
 declare -a exec_base=(exec --user "$DIND_USER")
@@ -393,10 +393,6 @@ docker_exec_env() {
   docker_exec env "${run_env[@]}" "$@"
 }
 
-docker_exec_root_env() {
-  docker exec "$DIND_NAME" env "${run_env[@]}" "$@"
-}
-
 run_setup=0
 case "$DIND_BOOTSTRAP" in
   always)
@@ -420,10 +416,7 @@ esac
 
 if [[ "$run_setup" == "1" ]]; then
   info "running scripts/setup.sh inside DinD"
-  docker_exec_root_env ./scripts/setup.sh
-  if [[ "$DIND_USER" != "root" ]]; then
-    docker exec "$DIND_NAME" chown -R "$DIND_USER" "$DIND_HOME_DIR"
-  fi
+  docker_exec_env ./scripts/setup.sh
 fi
 
 info "running scripts/run_fleet.sh ${fleet_args[*]} inside DinD"
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -430,12 +430,48 @@ if [[ -n "${CLAUDE_TGZ_SOURCE:-}" && -n "${CLAUDE_WHEEL_DIR_SOURCE:-}" ]]; then
 fi
 
 # ---- 7. Clone repo ----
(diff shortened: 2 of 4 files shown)


qiyuangong (approved):
LGTM

PR #61, opened 2026-07-29
Clean up Harbor analyzer when launcher exits

### LINKED ISSUE #54: Foreground and command-mode Harbor runs leak orphan analyzer processes when the launcher exits
## Summary

Harbor starts the handover-follow analyzer with `nohup setsid ... --follow`, so it is intentionally detached from the launcher and never exits on its own. An independent analyzer lifecycle supervisor is started only for `--detach` runs.

Foreground Zellij runs and the direct-command path rely on `start.sh` reaching their normal post-run cleanup code. If the controlling terminal disappears, the launcher receives HUP, or the launcher is otherwise terminated before that point, the analyzer survives as a PID-1 orphan indefinitely.

This is related to the broader status/stop work in #44, but it is a distinct automatic lifecycle-cleanup defect: no explicit user stop should be required after the benchmark has already reached a terminal state.

## Code path

`Agents/utils/common/Harbor/start.sh`:

- `harbor_start_analyzer_if_enabled()` launches the analyzer using `nohup setsid python3 ... --follow` (around line 240).
- `harbor_finish_analyzer_lifecycle()` correctly waits for monitor completion, drains pending handovers, and stops the analyzer (around lines 358-363).
- The direct-command path calls that cleanup only after its child returns normally (around line 537). It installs temporary `TERM`/`INT` handlers, but no global `EXIT`/`HUP` cleanup.
- The foreground Zellij path calls cleanup only after the Zellij command returns to the shell (around line 621), with no lifecycle supervisor and no global exit trap.
- `harbor_start_detached_analyzer_supervisor_if_enabled()` exists, but is invoked only inside the `DETACH_MODE=true` branch (around line 598).

The analyzer therefore has daemon semantics in every mode, but independent lifecycle ownership only in detached mode.

## Observed evidence

As of 2026-07-29 00:04 local time, three terminal runs still had live analyzer processes:

| Run | Benchmark/monitor terminal state | Analyzer state | Live analyzer |
| --- | --- | --- | --- |
| `2026-07-28-2050-harbor-tui` | benchmark `exit=0`; monitor `completed` | one failed handover reached 3 attempts and `retry_exhausted=true` | PID `3605777`, alive >3h |
| `2026-07-28-175523-tb21-analyzer-20` | benchmark `exit=0`; monitor had already terminated | two failed handovers reached 3 attempts and `retry_exhausted=true` | PID `2378553`, alive >6h |
| `2026-07-28-1628-harbor-tui` | benchmark `exit=35`; monitor terminal | processed state present, no pending retry | PID `4032939`, alive >7.5h |

For all three processes:

- PPID is `1`.
- The analyzer is its own session leader (`SID == PID`) because of `setsid`.
- CPU is idle and it is polling in `do_select`.
- `harbor_analyzer_pid_matches_run` returns true, so safety identity checks would allow cleanup.
- No analyzer supervisor PID/identity exists.
- The analyzer PID file remains present; `harbor_stop_analyzer()` would remove it, indicating the stop path was never reached.

The clean `2050` run is the strongest reproduction: the benchmark exited successfully, monitor reported `completed`, analyzer retries were exhausted, and no handover remained to drain, yet the analyzer continued polling for hours.

By contrast, an active `--detach` run (`2026-07-28-2209-harbor-tui`) has a live analyzer supervisor with a valid PID/start-time identity. This confirms the lifecycle ownership differs by launch branch.

## Impact

- Every interrupted foreground/command-mode run can leave another permanent polling process.
- Each observed analyzer retains roughly 20 MiB RSS; accumulation is unbounded.
- The analyzer process retains its API credential and proxy configuration in its environment.
- If an old handover file changes or a run directory is reused, a stale analyzer can invoke the model again and race with a new analyzer writing the same output directory.
- PID files continue to describe a live but logically finished service.

## Expected behavior

Every analyzer must have exactly one lifecycle owner independent of launcher/terminal survival. After the benchmark and monitor are terminal, pending handovers should drain within the configured budget and the analyzer must exit in all launch modes.

## Suggested fix

1. Start the analyzer lifecycle supervisor for every analyzer launch, not only `--detach`.
2. Make foreground and direct-command cleanup delegate to or coordinate with that same supervisor instead of maintaining separate ownership paths.
3. Add `EXIT`, `HUP`, `TERM`, and `INT` cleanup/fallback handling after analyzer startup.
4. Have the supervisor remove its own PID/identity files on natural exit, guarded by PID/start-time identity so it cannot remove a replacement supervisor's files.
5. Ensure a monitor crash while the benchmark is still live does not incorrectly look like successful benchmark completion; use explicit benchmark/monitor terminal evidence where possible.

An alternative is to make analyzer follow mode accept an explicit terminal sentinel and self-exit after its handover queue drains, reducing reliance on an external shell supervisor.

## Acceptance criteria

- Foreground Zellij normal completion leaves no analyzer process.
- Foreground launcher HUP/terminal loss leaves no analyzer process after the drain budget.
- Direct-command success, nonzero exit, `INT`, `TERM`, and `HUP` all clean up the analyzer.
- Detached completion continues to drain and stop correctly.
- Pending handovers are not lost during cleanup; exhausted retries do not keep the process alive.
- Supervisor/analyzer PID and identity files are removed after terminal cleanup.
- Unrelated or PID-reused processes remain protected by the existing PID/start-time and command-line identity checks.
- Integration tests exercise real child processes for all launch branches; current analyzer runtime tests cover drain-budget calculation but not process lifecycle.

0407nancy:
## Summary

- arm an `EXIT` fallback after successful analyzer startup in foreground and direct-command runs
- handle `HUP` in the existing direct-command signal path
- route foreground `HUP`, `INT`, and `TERM` through lifecycle cleanup
- leave detached analyzer supervision unchanged

## Root cause

The analyzer runs in its own detached session, while foreground and direct-command runs previously stopped it only when `start.sh` reached normal post-run cleanup. Launcher termination could therefore leave the analyzer polling after the run ended.

## Validation

- `git diff --check`
- `bash Agents/utils/common/Harbor/tests/test_start_user_feedback.sh`
- `bash Agents/utils/common/Harbor/tests/test_host_defaults.sh`
- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test_*.py'` (55 tests)
- focused Bash check that trapped `HUP` reaches the `EXIT` cleanup

Fixes https://github.com/sii-system/agent-fleet/issues/54

### DIFF
--- a/Agents/utils/common/Harbor/start.sh
+++ b/Agents/utils/common/Harbor/start.sh
@@ -508,13 +508,15 @@ if [[ $# -gt 0 ]]; then
       harbor_rollback_analyzer_startup
       exit 1
     fi
+    trap 'harbor_finish_analyzer_lifecycle' EXIT
   fi
   command_pid=""
   command_signal=""
   trap 'command_signal=TERM; [[ -z "$command_pid" ]] || kill -TERM "$command_pid" 2>/dev/null || true' TERM
   trap 'command_signal=INT; [[ -z "$command_pid" ]] || kill -INT "$command_pid" 2>/dev/null || true' INT
+  trap 'command_signal=HUP; [[ -z "$command_pid" ]] || kill -HUP "$command_pid" 2>/dev/null || true' HUP
   (
-    trap - TERM INT
+    trap - TERM INT HUP
     exec "$@"
   ) <&0 &
   command_pid="$!"
@@ -529,12 +531,14 @@ if [[ $# -gt 0 ]]; then
     command_rc="$?"
   fi
   set -e
-  trap - TERM INT
+  trap - TERM INT HUP
   if [[ -n "$command_signal" ]]; then
+    trap - EXIT
     harbor_rollback_analyzer_startup
     kill "-$command_signal" "$$"
   fi
   harbor_finish_analyzer_lifecycle
+  trap - EXIT
   exit "$command_rc"
 fi
 
@@ -613,12 +617,17 @@ if ! harbor_start_analyzer_if_enabled; then
   harbor_rollback_analyzer_startup
   exit 1
 fi
+trap 'harbor_finish_analyzer_lifecycle' EXIT
+trap 'exit 129' HUP
+trap 'exit 130' INT
+trap 'exit 143' TERM
 harbor_print_run_receipt
 zellij_status=0
 env -u ZELLIJ_SESSION_NAME zellij \
   --session "$ZELLIJ_SESSION_NAME" \
   --new-session-with-layout "$LAYOUT_FILE" || zellij_status="$?"
 harbor_finish_analyzer_lifecycle
+trap - EXIT HUP INT TERM
 final_status=0
 harbor_report_foreground_result "$zellij_status" || final_status="$?"
 exit "$final_status"
--- a/Agents/utils/common/Harbor/start.sh
+++ b/Agents/utils/common/Harbor/start.sh
@@ -618,14 +618,15 @@ if ! harbor_start_analyzer_if_enabled; then
   exit 1
 fi
 trap 'harbor_finish_analyzer_lifecycle' EXIT
-trap 'exit 129' HUP
-trap 'exit 130' INT
-trap 'exit 143' TERM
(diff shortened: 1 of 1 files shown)


qiyuangong (approved):
LGTM

PR #62, opened 2026-07-29
Refresh stale DinD home volumes

JasperChan929:
## 1. Why the change?

`DIND_BOOTSTRAP=missing` can treat a stale persistent home volume as fully
initialized when Pi, its model configuration, and the Harbor skill exist but
the managed prerequisite path contract does not. Because setup is skipped,
`$DIND_HOME_DIR/.config/agent-fleet/paths.env` is never created and later runs
can depend on coincidental system tool paths.

## 2. Summary of the change

- Require `paths.env` in the DinD `missing` bootstrap probe.
- Re-run setup when the existing home volume lacks that file.
- Preserve the existing fast path when every required artifact is present.
- Add positive and negative shell regression coverage for both decisions.

## 3. Other details

Validation:

- `env -i PATH="$PATH" HOME="$HOME" USER="${USER:-}" bash scripts/tests/test_dind_run.sh`
- `python3 -m unittest scripts/tests/test_setup.py`
- `bash -n scripts/dind-run.sh scripts/tests/test_dind_run.sh`
- `git diff --check`

### DIFF
--- a/scripts/dind-run.sh
+++ b/scripts/dind-run.sh
@@ -402,9 +402,10 @@ case "$DIND_BOOTSTRAP" in
     # Extract the version number instead of matching the whole output line:
     # setup.sh does the same, and an exact-line match would re-run setup on
     # every invocation if `pi --version` ever prints more than the bare version.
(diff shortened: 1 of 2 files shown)


qiyuangong (approved):
LGTM

Issue #56, opened 2026-07-29
Harbor monitor/analyzer pipeline robustness: unhandled exceptions kill the analyzer daemon, monitor JSON writes are not atomic, and partially-failed handovers re-run in full


JasperChan929:
## Summary

Code review of the monitor → analyzer pipeline found three related robustness defects. They share one theme: the pipeline behaves correctly on the happy path but degrades badly on the unhappy one.

1. Any exception other than `ValueError` kills the analyzer follow daemon; remaining handovers are silently never analyzed.
2. Every consumer-facing monitor JSON file is written non-atomically; concurrent readers can observe empty or truncated JSON.
3. A handover that publishes with per-task failures (exit code 3) is treated as unprocessed and re-run in full up to 3 times, re-analyzing already-successful tasks and re-publishing duplicate reports.

Items 1–2 are mechanical fixes. Item 3 needs a semantics decision first (see below).

## 1. Unhandled exceptions kill the analyzer follow daemon

`Agents/utils/common/Harbor/scripts/analyzer_subagent.py` (follow loop, around line 364) catches only `ValueError`:

- `ValueError` is what `load_json` and `validate_handover` raise (`harbor_analyzer/contract.py`, around line 174) — the handler was designed for input-contract failures.
- Everything else escapes to `main()` and kills the daemon: `OSError` from any of the many artifact writes in `harbor_analyzer/runner.py` / `pi.py` (disk-full is realistic — see #53, where a single task analysis wrote ~842 MB of event log), and with `--max-concurrency > 1`, any worker-thread exception re-raised by `future.result()` (`harbor_analyzer/runner.py`, around line 788).

Consequences when it fires:

- The in-flight handover is not recorded in follow state (`_record_follow_failure` never runs), and all queued handovers are never analyzed.
- In foreground/command modes there is no supervisor to restart it (#54).
- At run end, `harbor_wait_for_analyzer_drain` in `start.sh` sees the dead PID and returns as "drained" — the loss is silent.

Suggested fix: broaden the per-handover handler to `except Exception`, reusing the existing record-failure/backoff/continue path, and log the traceback. The loop's designed invariant (one handover failure must not kill the daemon) already exists; it just doesn't cover all failure modes.

## 2. Monitor JSON outputs are written non-atomically

[...]

PR #65, opened 2026-07-29
Treat degraded analyzer handovers as complete

0407nancy:
## Summary

- treat analyzer exit code `3` as a completed handover with task-level failures
- persist the handover in follow state instead of retrying its full task list
- retain the existing per-task fallback reports and internal Pi retry behavior

## Root cause

`run_handover` returns `3` only after publishing the handover artifacts with fallback reports for failed task analyses. Follow mode previously treated every nonzero code as retryable, so it reran successful tasks and created duplicate publications.

## Validation

- `python3 -m unittest Agents.utils.common.Harbor.tests.test_harbor_analyzer_runtime` (21 tests)
- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test_*.py'` (56 tests)
- `python3 -m py_compile Agents/utils/common/Harbor/scripts/analyzer_subagent.py Agents/utils/common/Harbor/tests/test_harbor_analyzer_runtime.py`
- `git diff --check`

Addresses the whole-handover retry item in:

- https://github.com/sii-system/agent-fleet/issues/56
- https://github.com/sii-system/tasks/issues/150#issuecomment-5112305951

### DIFF
--- a/Agents/utils/common/Harbor/scripts/analyzer_subagent.py
+++ b/Agents/utils/common/Harbor/scripts/analyzer_subagent.py
@@ -377,7 +377,7 @@ def main() -> int:
                 )
                 _save_follow_state(state_path, processed=processed, failed=failed)
                 continue
-            if exit_code == 0:
+            if exit_code in {0, 3}:
                 processed.add(follow_key)
                 failed.pop(follow_key, None)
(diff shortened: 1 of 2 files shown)


qiyuangong (approved):
LGTM

PR #66, opened 2026-07-29
Retain whole-run analyzer publication manifest

0407nancy:
## Summary

- retain every handover's immutable artifact publication in `analyzer-artifacts-latest.json`
- replace an older entry when the same handover is published again
- update the whole-run publication list under the existing publish lock
- preserve the top-level latest-publication fields for compatibility

## Root cause

Each handover previously replaced the complete latest manifest. Multi-handover runs therefore exposed only the final handover's artifact paths even though earlier immutable publications remained on disk.

The manifest now includes a `publications` list for the current run. Benchmark reports, env/infra task files, and fix-line indexes remain immutable per-handover artifacts and are not rewritten into a second aggregate format.

## Validation

- `python3 -m unittest Agents.utils.common.Harbor.tests.test_harbor_analyzer_runtime` (22 tests)
- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test_*.py'` (57 tests)
- `python3 -m py_compile Agents/utils/common/Harbor/scripts/harbor_analyzer/runner.py Agents/utils/common/Harbor/tests/test_harbor_analyzer_runtime.py`
- `git diff --check`

Addresses the incomplete whole-run latest manifest item in:

- https://github.com/sii-system/agent-fleet/issues/56
- https://github.com/sii-system/tasks/issues/150#issuecomment-5112305951

### DIFF
--- a/Agents/utils/common/Harbor/scripts/harbor_analyzer/runner.py
+++ b/Agents/utils/common/Harbor/scripts/harbor_analyzer/runner.py
@@ -13,7 +13,7 @@ from uuid import uuid4
 
 from . import PROMPT_VERSION, SCHEMA_VERSION, TAXONOMY_VERSION
 from .contract import validate_handover
-from .io import stable_hash, utc_now, write_json_atomic, write_text_atomic
+from .io import load_json, stable_hash, utc_now, write_json_atomic, write_text_atomic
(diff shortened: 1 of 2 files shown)


qiyuangong (approved):
LGTM

PR #68, opened 2026-07-29
Enable cgroup v2 nesting in DinD

JasperChan929:
## 1. Why the change?

On cgroup v2 hosts, the custom privileged DinD runner could start its own
daemon but failed when Harbor launched a nested task container:

```text
cannot enter cgroupv2 "/sys/fs/cgroup/docker" with domain controllers -- it is in an invalid state
```

The DinD entrypoint left its own processes in the delegated root cgroup while
nested dockerd attempted to distribute domain controllers to child cgroups.
As a result, every Docker-backed benchmark task failed during environment
setup before the agent could run.

## 2. Summary of the change

- Move DinD's root cgroup processes into an `/init` child before dockerd starts.
- Enable all delegated cgroup v2 controllers for nested containers, retrying
  around `docker exec` process races.
- Keep cgroup v1 behavior unchanged.
- Include the new helper in the DinD image and image fingerprint so existing
  runners rebuild onto the fixed image.
- Add focused cgroup v1/v2 tests and run them in PR CI.

## 3. Other details

Validation:

- `python3 -m unittest discover -s scripts/tests -v` — 116 tests passed.
- `bash scripts/tests/test_dind_cgroup_v2.sh`
- Shell syntax checks and `git diff --check`
- Built the updated DinD image on a cgroup v2 host and started a real nested
  `busybox` container successfully. The DinD root cgroup had zero processes,
  its own processes were under `/init`, and all delegated controllers were
  enabled.

The real validation used no API credentials or other secrets.

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -42,5 +42,6 @@ jobs:
           set -euo pipefail
           python3 -m unittest discover -s tests -v
           python3 -m unittest discover -s scripts/tests -v
+          bash scripts/tests/test_dind_cgroup_v2.sh
           bash scripts/tests/test_dind_run.sh
           python3 -m unittest discover -s .github/scripts/tests -v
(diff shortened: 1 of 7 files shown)


qiyuangong (approved):
LGTM

PR #69, opened 2026-07-30
Simplify interactive Opik setup

JasperChan929:
## 1. Why the change?

Interactive setup currently exposes `TRACE_TO_OPIK` as a separate yes/no decision before asking for `OPIK_URL`. These values are operationally related, so the extra tracing question adds user-facing complexity without adding a useful first-run choice.

The internal switch still needs to remain authoritative because it controls Harbor, OpenClaw, PinchBench, and ClawBio behavior and must support temporarily disabling Opik while retaining a saved endpoint.

## 2. Summary of the change

- Replace the tracing yes/no prompt with one optional `OPIK_URL` prompt.
- Derive and persist `TRACE_TO_OPIK=true` when a URL is supplied and `false` when it is left empty.
- Preserve explicit `TRACE_TO_OPIK` overrides, legacy URL-only configurations, and saved Opik settings while tracing is disabled.
- Update setup documentation and the public configuration template to distinguish the simple interactive flow from the advanced fleet-wide override.
- Add regression coverage for URL entry, URL omission, legacy configuration, and explicit trace-off preservation.

## 3. Other details

Validation:

- `python3 -m unittest discover -s scripts/tests` — 118 tests passed
- `bash scripts/tests/test_dind_cgroup_v2.sh`
- `bash scripts/tests/test_prerequisites.sh`
- `bash -n scripts/setup.sh`
- `git diff --check`
- Real TTY smoke tests in an isolated temporary HOME/repository:
  - empty URL persisted `TRACE_TO_OPIK=false`
  - entered URL persisted `TRACE_TO_OPIK=true`, the URL, and the default workspace
  - explicit `TRACE_TO_OPIK=false` retained the saved URL
  - each run completed the real Docker daemon permission check

### DIFF
--- a/config.env
+++ b/config.env
@@ -24,14 +24,9 @@
 # API_KEY=sk-your-key
 # MODEL=your-model-id
 
-# ── Opik tracing ─────────────────────────────────────────────────────
-# setup.sh asks only for this optional endpoint, then records an explicit
-# internal TRACE_TO_OPIK state in config.local.env.
+# ── Opik (optional) ──────────────────────────────────────────────────
+# setup.sh asks for this endpoint and persists whether Opik should be used.
 # OPIK_URL=https://opik.example.com/api
-# Advanced fleet-wide override: false disables every supported benchmark path
-# while retaining any saved endpoint and overrides stale OPIK_PLUGIN=enabled
-# settings. true requires OPIK_URL.
-# TRACE_TO_OPIK=true
 # OPIK_PROJECT_NAME=
 
 # ── Package mirrors ───────────────────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


config.env
```diff
@@ -25,12 +25,13 @@
...
-# setup.sh records an explicit choice in config.local.env. Uncomment only when
-# configuring the repository without setup.
-# TRACE_TO_OPIK=true
+# setup.sh asks only for this optional endpoint, then records an explicit
+# internal TRACE_TO_OPIK state in config.local.env.
 # OPIK_URL=https://opik.example.com/api
+# Advanced fleet-wide override: false disables every supported benchmark path
+# while retaining any saved endpoint and overrides stale OPIK_PLUGIN=enabled
+# settings. true requires OPIK_URL.
+# TRACE_TO_OPIK=true
```
qiyuangong: Can we remove `TRACE_TO_OPIK` from README? User don't have to know this env

qiyuangong (approved):
LGTM

PR #71, opened 2026-07-30
Unify fleet configuration precedence

JasperChan929:
## 1. Why the change?

Fleet entry points loaded `config.env` and `config.local.env` independently. The duplicated logic made source precedence easy to drift between direct runs, DinD, prompt mode, and Harbor, and tool-specific variables could be promoted into global model configuration outside the tool that owns them.

This change establishes one deterministic contract for same-name configuration values: runtime/exported environment, then `config.local.env`, then `config.env`, then subsystem defaults.

Related context: https://github.com/sii-system/tasks/issues/150#issuecomment-5114350786

## 2. Summary of the change

- Add a shared, source-only repository configuration loader and load each repository root once per process chain.
- Use the loader from `run_fleet.sh`, `dind-run.sh`, prompt mode, and direct Harbor entry points.
- Keep `TB_MODEL` and `ANTHROPIC_*` compatibility variables scoped to Harbor/Claude Code fields instead of rewriting global `MODEL`, `BASE_URL`, or `API_KEY`.
- Add unit and integration coverage for runtime/local/public precedence, tool-alias isolation, direct Harbor behavior, and DinD credential handoff.

## 3. Other details

This PR intentionally does not validate endpoint/model/key correctness, add model preflight requests, change Analyzer diagnosis, or add a run-level circuit breaker. Those concerns remain separate from configuration source resolution.

Validation:

- `python3 -m unittest discover -s scripts/tests` — 123 tests passed
- `python3 -m unittest discover -s Agents/utils/common/Harbor/tests -p 'test_*.py'` — 57 tests passed
- All shell tests under `scripts/tests` and `Agents/utils/common/Harbor/tests` passed
- `bash -n` and `git diff --check` passed

All committed test credentials use clearly fake placeholders.


qiyuangong:
Basic design looks good. PR is mergable after fixing P1 and P2 issues.

qiyuangong (approved):
LGTM

PR #80, opened 2026-08-03
refactor(harbor): split monitor and controller control flow

0407nancy:
## Summary

- separate Harbor observation/evaluation from controller policy and execution while preserving the existing runner entry point
- keep wait and notify behavior distinct from restart/stop decisions, with restart and stop exposed as user-selectable actions instead of automatic commands
- isolate analyzer dispatch behind the controller boundary and document the minimal-change structure
- add controller unit coverage and monitor/controller integration coverage

## Validation

- Harbor unittest suite: 81 tests passed
- real benchmark run exercised the monitor follow flow and wait/notify decision paths
- verified restart commands are not automatically triggered by the refactored flow


qiyuangong:
Need to fix Ruff issues.

qiyuangong (approved):
LGTM

PR #99, opened 2026-08-17
feat(harbor): include Fixer report in benchmark summary

0407nancy:
## 1. Why the change?

The human-readable benchmark summary currently covers Monitor and Analyzer results only. When Fixer produces fix-report-latest.md, users must open that full report separately to understand the fix outcome.

## 2. Summary of the change

- Accept the Fixer Markdown report as an optional benchmark-summary input.
- Append only the report summary, applied changes, and remaining issues.
- Keep verification statistics in the summary table and omit the duplicate verification subsection and execution details.
- Link the full Fixer report and show a clear message when it has not been generated.
- Pass the conventional OUTPUT_PATH/fixer/fix-report-latest.md location from the Harbor launcher.
- Support both the agreed section names and the current experimental Reporter headings without reading Fixer JSON artifacts.

## 3. Other details

This does not start or control Fixer and does not add Controller UI. The summary writer remains backward-compatible for callers that do not provide a Fixer report path.

Validation:
- 11 benchmark summary tests passed.
- 140 sandbox-compatible Harbor tests passed; the one loopback-socket test passed outside the sandbox.
- Ruff checks and bash syntax checks passed.


qiyuangong (approved):
LGTM

PR #103, opened 2026-08-17
feat(harbor): add controller fixer workflow

0407nancy:
## 1. Why the change?

Harbor Fixer can plan, policy-check, execute, and verify fixes, but those stages did not have one run-local Controller workflow. Users could not review the exact approved plan before mutation or follow the same workflow through verification and the human-readable result.

## 2. Summary of the change

- Add Controller commands to start, approve, cancel, reset, and inspect a run-local Fixer workflow.
- Keep `fixer start` non-mutating: it generates the Fix Plan, runs policy, and presents the exact plan for review.
- Bind approval to the run, workflow, request, and Fix Plan SHA-256; reject changed or denied plans.
- After approval, automatically execute the plan, call the merged Fixer smoke-verification workflow, write `fix-report-latest.md`, and replace only the existing Fixer section in `benchmark-summary.md`.
- Expose `executing`, `verifying`, and `reporting` states plus execution counts, verification outcome, report status, and artifact paths through Controller status.
- Prevent a second workflow or reset while tracked Fixer Pi, action, or verification work may still be active; wait for action process groups before publishing completion.
- Require the final Analyzer handoff to be drained before starting Fixer.

## 3. Other details

- Approval covers execution followed by smoke verification and report generation. Verification and reporting do not introduce additional user decisions or Controller commands.
- The verification rerun uses the repository `start.sh`; users do not need to configure restart, stop, or verification commands.
- Cancellation is cooperative before execution and rejects a plan awaiting approval. It does not claim to stop execution or verification after mutation starts.
- Verification and reporting failures are recorded at their actual stage and returned to the caller. No retry, fallback report, force reset, or marker-expiry behavior was added.
- Fixer results remain in `$RUN_DIR/fixer`; reset clears only Controller-owned control state when no tracked process remains.
- Tests: 65 focused Controller/Fixer/verification/report/summary tests passed, and Ruff 0.16.0 passed. The full Harbor suite ran 309 tests: 308 passed; the only import error is an unrelated latest-main QZ module using `typing.override` under the local Python 3.10 runtime.


qiyuangong (approved):
LGTM

PR #110, opened 2026-08-18
Rename legacy TB environment variables

qiyuangong:
## 1. Why the change?

`TB_*` environment variables are legacy names from when Harbor only supported Terminal Bench. They now describe generic Harbor runtime, task, dataset, rollout, and tracing behavior, and the names no longer match their actual scope.

This also aligns Agent Fleet's Opik metadata producers with [agent-opik-plugin#13](https://github.com/sii-system/agent-opik-plugin/pull/13), which consumes the `HARBOR_*` metadata contract.

## 2. Summary of the change

- Rename generic `TB_*` environment variables and internal variables to `HARBOR_*` across Harbor runners, model-fusion paths, RL workers, fleet scripts, tests, docs, and skills.
- Rename Opik tags from `tb-{run,task,trial}:` to `harbor-{run,task,trial}:` and forward `HARBOR_DATASET` alongside the other Harbor metadata fields.
- Consolidate duplicate configuration onto the existing canonical names `DATASET_PATH`, `AGENT`, and `TRACE_TO_OPIK`.
- Remove obsolete TerminalBench-only dataset auto-cloning with an explicit local-dataset migration error, plus obsolete version and unused wheel-port variables.
- Pin the tracing-plugin submodule to agent-opik-plugin#13's renamed contract and keep the retired RL template request spelling fail-closed without supporting it as an alias.
- Migrate persisted offline Claude package settings from the legacy managed `~/.bashrc` block with a visible warning.
- Add a repository invariant test that rejects reintroduction of legacy generic names.

True TerminalBench identifiers such as `tb21`, `run_tb21.sh`, and `tb_tasks.txt` remain unchanged. Compatibility-sensitive `/opt/tb-*` filesystem paths are also retained.

## 3. Other details

This is an environment-variable migration. Operator configuration and automation using the removed `TB_*` names must move to their `HARBOR_*` replacements, or to `DATASET_PATH`, `AGENT`, and `TRACE_TO_OPIK` where those contracts were consolidated.

Validation:

- `python3 -m unittest tests.test_harbor_env_naming -v`
- Ruff on all changed Python files
- `bash -n` on all changed shell files
- Python byte-compilation and `git diff --check`
- Broad affected pytest suites: 456 passed; 15 pre-existing macOS path/Bash compatibility failures were reproduced unchanged on the untouched base commit
- Fleet batch regression suite: 16 passed, including preservation of exported `HARBOR_RUNS`
- Pinned agent-opik-plugin suite: 241 passed


LongHanyu (approved):
LGTM

PR #124, opened 2026-08-19
feat(harbor): expose fixer report CLI

AmaoLearning:
## 1. Why the change?

Harbor Fixer can generate machine-readable and Markdown reports through its Python APIs, but operators need a standalone CLI path to regenerate reports from persisted verification and Analyzer artifacts without rerunning planning, execution, or verification.

## 2. Summary of the change

- Add report-only CLI arguments for the verification result, optional baseline run, and baseline monitor policy.
- Route report-only mode through the existing report generation API and print the report-agent prompt when requested.
- Isolate report-only parsing from inherited planning, execution, and verification settings that this mode does not use.
- Document the report-only workflow and cover its CLI validation behavior.

## 3. Other details

- Rebased onto the current upstream main, including #122.
- Full Harbor Fixer suite: 73 tests passed.
- Ruff 0.16.0 passes for the changed Python files; upstream main Ruff CI is green.

### DIFF
--- a/Agents/utils/common/Harbor/README.md
+++ b/Agents/utils/common/Harbor/README.md
@@ -428,13 +428,28 @@ Verification writes
 `verification-result-latest.json`. If the Fix Plan was generated without a
 local monitor, select `claude-code`, `opencode`, or `oracle` with `--agent`.
 
-### Human-readable Fixer result
+### Generate a report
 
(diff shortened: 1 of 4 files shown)


qiyuangong (approved):
LGTM

PR #140, opened 2026-08-24
fix(opencode): validate Node runtime before install

JasperChan929:
## 1. Why the change?

Some SWE-Rebench final images expose an npm command or shim without a working Node.js runtime. The OpenCode installer treated command presence as readiness, skipped the prepared Node runtime, and then failed during agent setup with `Could not determine Node.js install directory`.

The OpenCode 401 seen on the older benchmark commit was already addressed by PR #132. This PR does not add another credential-delivery mechanism.

## 2. Summary of the change

- Add one functional Node/npm readiness check that requires both commands and successful `node --version` / `npm --version` execution.
- Use that check consistently before the cached runtime, Node distribution download, package-manager fallback, and final OpenCode installation.
- Activate candidate runtimes through one helper that refreshes the shell command cache, validates both tools, and removes failed symlinks before fallback.
- Prefer a validated `/usr/bin` runtime after package-manager installation so broken shims earlier on PATH cannot continue shadowing healthy binaries.
- Add behavioral coverage for missing Node with broken npm, broken Node, healthy runtimes, and post-symlink validation.

## 3. Other details

Scope is limited to the OpenCode Node/npm installer and its tests. There are no mirror-index, image-push, Template-creation, or new secret-injection changes.

Validation:

- OpenCode tests: 20 passed, 1 skipped because the optional Harbor dependency is not installed locally.
- generation config tests: 16 passed, including repeated OpenCode secret loading.
- QZ environment tests: 42 passed.
- QZ shell test, Python compile, diff check, and Ruff 0.16.0: passed.
- Linux targeted Node installer tests: 5 passed.
- Latest-main QZ auth canary: no 401; agent_result recorded 7040 input and 7 output tokens.
- QZ Node canary on the existing `hypothesisworks__hypothesis-2649` ready Template: OpenCode setup completed successfully with no Node install error or 401. The subsequent model request reached the gateway and returned an unrelated server error.


qiyuangong (approved):
LGTM

PR #150, opened 2026-08-31
feat(opensandbox): cache prebuild uploads locally

LJXLJXLJX:
## 1. Why the change?

Large datasets may contain tens of thousands of tasks. When prebuild is restarted, even if only one task is missing, every task must recalculate its hash and perform a Registry request to discover the cache hit.

This makes resuming prebuild unnecessarily slow.

## 2. Summary of the change

- Cache successfully uploaded Bundle records locally.
- Reuse matching local records without querying the Registry.
- Add an optional fast-resume mode that skips task hash verification and trusts the local uploaded-task record.
- Fall back to the existing Registry lookup and build flow when no local record exists.
- Avoid unnecessary package-source health requests for local hits in fast-resume mode.

## 3. Other details

Hash verification remains enabled by default. Fast-resume mode is intended for stable datasets where task content rarely changes and Registry artifacts are retained.


qiyuangong (approved):
LGTM

PR #158, opened 2026-09-02
fix(harbor): keep runner Python version consistent

JasperChan929:
## Summary

- make HARBOR_RUNNER_PYTHON_VERSION the canonical runner Python version setting
- keep HARBOR_RUNNER_PYTHON as a compatible legacy alias and reject conflicting values
- persist an explicitly selected host-runner Python version so setup and later runtime validation use the same contract
- cover canonical, legacy, conflict, marker, and setup-to-runtime behavior

## Why

A one-shot setup such as HARBOR_RUNNER_PYTHON=3.12.3 ./scripts/setup.sh could create a valid runner, but a later Harbor invocation defaulted its validator to 3.12.13. That made the runner fail validation even though setup had succeeded.

The host runner now records an explicitly requested version in .agent-fleet-python-version. Runtime selection uses the canonical variable first, then the legacy alias, then that marker, and finally the pinned default. Image-provided runners do not consume the host marker.

## Validation

- bash Agents/utils/common/Harbor/tests/test_setup_runner_env.sh
- Python 3.12.3: python -m unittest discover -s Agents/utils/common/Harbor/tests -p "test_*.py" (563 passed)
- Python 3.12.3: python -m unittest scripts.tests.test_setup scripts.tests.test_setup_config (25 passed)
- bash -n on changed shell scripts
- git diff --check

Context: sii-system/tasks#168


qiyuangong (approved):
LGTM

PR #168, opened 2026-09-03
fix: retry transient skopeo copy failures

LJXLJXLJX:
## 1. Why the change?

OpenSandbox image publishing occasionally fails due to transient network errors, such as `connection reset by peer` during `skopeo copy`.

## 2. Summary of the change

Retry `skopeo copy` up to 3 total attempts with a 3-second delay. The first two failures are logged as warnings, while the final failure is raised as an error.

## 3. Other details

The retry applies only to the image copy step. Build, login, inspection, and digest verification behavior remains unchanged.


LongHanyu (approved):
LGTM

PR #176, opened 2026-09-07
fix(web search): fix timeout, retry, and PDF handling for Claude agent

LongHanyu:
## 1. Why the change?

Some bugs are found when using Claude agent do web searching

## 2. Summary of the change 

fix timeout, retry, and PDF handling for Claude agent


LJXLJXLJX (approved):
lgtm

PR #181, opened 2026-09-08
feat(ci): add no-publish Pi finding verification replay

qiyuangong:
## 1. Why the change?

Pi now has exact PR-head source through #180, but its findings still lack an independent factual check. Add a no-publish replay path to evaluate a verifier against known findings before changing live publication behavior.

For example, the replay fixtures cover [PR #178's false missing-module claim](https://github.com/sii-system/agent-fleet/pull/178#pullrequestreview-5136362667), plus [PR #179's real test/implementation mismatch](https://github.com/sii-system/agent-fleet/pull/179#discussion_r3953693173) and the same claim after its fix.

## 2. Summary of the change

- Add a CLI accepting one frozen candidate, canonical repository, base/head SHAs, and bounded source excerpt ranges. Read source through disposable Git stores using the existing bounded fetch, with anonymous access when no GitHub token is supplied; keep expected verdicts and label notes out of model input.
- Run a separate Pi verification session without tools, extensions, skills, or context-file discovery. Require a verdict, failure scenario, introduced behavior, counterevidence, and source citations.
- Check exact quotes, line ranges, source identities, and absence claims against Git. Confirmation requires both revisions and a cited changed file; rejection requires head evidence. Invalid citations become `insufficient_evidence`, with the original model verdict retained.
- Write private local artifacts with source/blob IDs, prompt/code/input hashes, model, timing, verdicts, and validation errors. Redact credentials in free-text values while preserving keys, schema enums, source IDs, and hashes, including with short dummy API keys. Make the 2000-character output-field limit explicit to the verifier and format-repair pass. Distinguish failed runs from rejected or uncertain findings. Add three historical seed cases and an operator runbook.

## 3. Other details

Validation:

- `PYTHONPATH=. python3 -m unittest discover -s .github/scripts/tests -q` — 324 tests passed.
- `uvx ruff@0.16.0 check --config .github/ruff.toml .github/scripts/pi_review_replay.py .github/scripts/tests/test_pi_review_replay.py .github/scripts/pi_pr_review.py .github/scripts/tests/test_pi_pr_review.py` — passed.
- `git diff --cached --check` — passed.
- End-to-end smoke on macOS with the workflow-pinned Pi 0.81.1, anonymous GitHub source transport from a fresh shallow clone, a one-character dummy model API key, and a local SSE provider stub: all three historical cases traversed the CLI successfully, preserved artifact status/verdict fields, produced the expected stub verdicts, and exposed zero tools in provider requests. A separate anonymous historical-head fetch also verified that trusted Git metadata and objects remain unchanged and temporary source storage is removed. This validates transport and evidence handling, not model quality.

Scope and limits: the live reviewer workflow and publication behavior are unchanged; this is an opt-in operator CLI with no GitHub publication calls. Context is explicitly selected and bounded, so missing callers/contracts must yield uncertainty. Citation integrity does not prove causal reasoning. The three fixtures are regression seeds, not a precision benchmark. Live model replay remains unmeasured because review-provider credentials were unavailable in the workspace. No migration is required.

### DIFF
--- a/.github/scripts/pi_pr_review.py
+++ b/.github/scripts/pi_pr_review.py
@@ -465,16 +465,20 @@ class PiClient:
             repository = github.api_root.removeprefix("https://api.github.com/repos/")
             if not re.fullmatch(r"[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+", repository):
                 raise PiReviewError("invalid GitHub source repository")
-            authorization = base64.b64encode(
-                f"x-access-token:{github.token}".encode()
-            ).decode()
             environment = dict(os.environ)
             environment.update({
                 "GIT_TERMINAL_PROMPT": "0",
-                "GIT_CONFIG_COUNT": "1",
-                "GIT_CONFIG_KEY_0": "http.https://github.com/.extraheader",
-                "GIT_CONFIG_VALUE_0": f"AUTHORIZATION: basic {authorization}",
+                "GIT_CONFIG_COUNT": "0",
             })
+            if github.token:
+                authorization = base64.b64encode(
+                    f"x-access-token:{github.token}".encode()
+                ).decode()
+                environment.update({
+                    "GIT_CONFIG_COUNT": "1",
+                    "GIT_CONFIG_KEY_0": "http.https://github.com/.extraheader",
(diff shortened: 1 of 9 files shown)


qiyuangong:
@codex review

PR #182, opened 2026-09-08
feat(opensandbox): add Dart and Julia package sources

LJXLJXLJX:
## 1. Why the change?

Fix image-build network timeouts affecting Dart and Julia tasks in SWE-rebench-v2.

## 2. Summary of the change

Add configurable, provider-neutral package sources for Dart Pub and Julia. The sources can point to a third-party mirror or cache gateway and are injected only during image builds.

## 3. Other details

Existing behavior is preserved when the new source settings are not configured.


LongHanyu (approved):
LGTM

PR #183, opened 2026-09-09
fix(opensandbox): detect parenthesized apt commands

LJXLJXLJX:
## 1. Why the change?

Parenthesized APT commands such as `RUN ( apt-get update ... )` were not detected, so OpenSandbox builds skipped APT mirror setup and could time out.

## 2. Summary of the change

- Detect parenthesized APT commands.
- Add a regression test covering setup and cleanup injection.

## 3. Other details

All 74 related unit tests and Ruff checks pass.


LJXLJXLJX:
On hold for now.

The APT renderer is accumulating shell-syntax-specific patches because it has to infer APT usage and source mutations from arbitrary Dockerfiles. Unlike PyPI, npm, or Go, APT has no single registry argument; its configuration is spread across source files and may change during the build. I’d like to pause here and reassess the problem boundary before adding more incremental complexity.

This is not a hard merge block: if the current fix is needed for affected builds, this PR can still be merged, and targeted follow-up patches can continue.

PR #188, opened 2026-09-10
refactor(opensandbox): intercept apt through runtime PATH

LJXLJXLJX:
## 1. Why the change?

Static Dockerfile analysis cannot reliably detect APT usage in nested shells, downloaded scripts, or subprocesses, and requires complex source mutation tracking and cleanup.

## 2. Summary of the change

- Replaced static APT analysis with build-time PATH wrappers for shell-form `RUN`.
- Rewrites invocation-time sources in an ephemeral shadow tree without modifying `/etc/apt`.
- Reconciles successful APT updates back to the original source identity.
- Preserves RUN options such as `--device` and keeps Git mirror mounts independent.
- Fixed image identity to use only the static task environment hash. Renderer/runtime mutations are documented as forbidden identity inputs and P0 bugs.

## 3. Other details

JSON exec-form APT calls, absolute `/usr/bin/apt*` paths, and custom libapt frontends intentionally bypass interception. Runtime assets are BuildKit-only and do not enter the final image.


LJXLJXLJX:
Closing this PR because we found a lower-level LLB-based approach that can cover all RUN forms consistently. We'll pursue that design instead.

PR #190, opened 2026-09-11
Fix SWE-rebench-v2 builds for migrated upstream repositories

LJXLJXLJX:
## 1. Why the change?

  Several SWE-rebench-v2 tasks reference GitHub repositories that have moved without redirects. Their original repository URLs now return 404, causing environment builds to fail before task setup begins.

  The canonical Pebble and SQLPad repositories retain the exact commits required by the dataset, so these tasks can be recovered without changing their baseline or evaluation semantics.

  ## 2. Summary of the change

  - Adds commit-scoped repository overrides:
      - petermattis/pebble → cockroachdb/pebble
      - rickbergfalk/sqlpad → sqlpad/sqlpad

  - Applies an override only to dataset commits verified to exist in the canonical repository.
  - Preserves the original repository name, commit, checkout directory, and task metadata.


LongHanyu (approved):
lgtm

PR #201, opened 2026-09-18
fix(ci): prepare pinned Harbor runner for E2E validation

qiyuangong:
## 1. Why the change?

[Harbor E2E run 35277982405](https://github.com/sii-system/agent-fleet/actions/runs/35277982405) stopped before any trials ran because the shared runner had `yicloud-sdk-python==0.3.1`, while `runner-requirements.txt` requires `0.4.1`. E2E prerequisites checked host tools and Python but never prepared the pinned Harbor runner.

## 2. Summary of the change

Prepare a workflow-owned Harbor runner under `RUNNER_TEMP` using the existing `setup_runner_env.sh`, matching the sampling nightly's approach. Export its paths through `GITHUB_ENV` so benchmark startup uses the validated dependencies. Setup failures stop the job before launch; shared host and image-provided runners remain untouched.

Add a regression test covering setup path selection, successful exports, and failure propagation, and document dependency preparation for both nightlies.

## 3. Other details

Validation:
- CI-helper suite: 370 tests, one skipped.
- Runner validation suite: three tests passed.
- Runner setup shell regression suite passed.
- Ruff 0.16.0 passed for the changed Python file.
- Workflow YAML parsed; `bash -n` passed for all 15 inline scripts; `git diff --check` passed.

A live self-hosted benchmark was not run. After merge, dispatch a new E2E canary on the updated branch to verify the host integration.


qiyuangong:
@codex review

PR #202, opened 2026-09-18
fix(opensandbox): stabilize base-image resolution and oracle patching

LJXLJXLJX:
## 1. Why the change?

Generated task artifacts currently depend on the registry URL used during generation, which can cause their contents to drift between environments.

Separately, stale Git index metadata in prebuilt images causes the oracle’s gold patch application to fail with `does not match index`, making the affected tasks receive zero reward.

## 2. Summary of the change

- Keep adapter-generated Dockerfiles registry-neutral.
- Resolve logical base images at build time through the configured registry and pin them to immutable digests using BuildKit named contexts.
- Refresh and validate the Git index before applying the oracle solution patch.

## 3. Other details

The first change keeps task artifacts portable across registry environments. The second prevents prebuilt-image index metadata from causing complete oracle failure.

Validated with the target image and 36 SWE-rebench tests.


LJXLJXLJX:
@codex review

qiyuangong:
@codex review

qiyuangong:
@codex review

LJXLJXLJX:
@codex review

qiyuangong (approved):
LGTM

PR #204, opened 2026-09-20
Fix/verifier bundle portability

LJXLJXLJX:
## 1. Why the change?

SWE-rebench-V2 verifier tasks require a sufficiently new Python, but many task images have no Python or an incompatible version. The previous bundle reused a dynamically linked runner Python and therefore depended on the task image's glibc, loader, and native ABI. This caused verifier startup failures before the task could be graded.

## 2. Summary of the change

- Add a pinned, SHA-verified CPython 3.12.14 x86_64 static-musl runtime.
- Validate the ELF and reject dynamic-loader or dynamic-library dependencies.
- Separate the reusable verifier runtime from the SWE-rebench-V2-specific bundle preparer and parser self-check.
- Mount the bundle read-only, extract and self-check it at Sandbox bootstrap, then run the verifier through the bundled Python. Runtime failures stop the verifier instead of falling back to task-image Python.

## 3. Other details

Ruff, `git diff --check`, and the bundle/preparer tests pass under both the system Python and the Harbor runner Python. A real OpenSandbox smoke test is still a separate validation step.


LJXLJXLJX:
commit tree is polluted, will recreate another PR

PR #205, opened 2026-09-20
feat(benchmark): add single-process high-concurrency benchmarks

LongHanyu:
## 1. Why the change?

High concurrency memory use too much

## 2. Summary of the change 

Reuse Harbor's native trial scheduler behind an opt-in mode to avoid one Python runner and log follower per Fleet worker. Keep one monitor pane and one batch worker, preserving exact task selection, per-task prebuilt images, and native result artifacts.

Isolate long OpenSandbox commands in a bounded shared executor so they cannot exhaust the default pool needed for control and file I/O. Keep the existing queue mode unchanged and reject dynamic RL in this first benchmark-only mode.


qiyuangong (approved):
LGTM

qiyuangong:
@codex review

LongHanyu:
Investigated the failing check and reran only the failed job (attempt 2): https://github.com/sii-system/agent-fleet/actions/runs/35499063347/job/106229216337

The failing job is **Self-hosted review**, not the code validation jobs. **Portable tests**, **Ruff lint**, and **Self-hosted summary** are passing.

Attempt 2 passed checkout and the Pi version check, then spent approximately 906 seconds in the review step before reporting `pi PR review failed: all review lenses failed`. The reviewer has a 900-second timeout, so timeout is a possibility, but the current logs do not establish the underlying cause: `pi_pr_review.py` catches each `PiReviewError` / `ModelResponseError` without reporting the per-lens exception, then emits only the aggregate failure.

This workflow executes the trusted base revision (`807e3dc`), not reviewer code from this PR. Changing the concurrency implementation or adding a diagnostic change only to the PR head would not repair this reviewer execution. No product-code changes or check bypasses were made.

The check is still failing. The next actionable step is for the CI maintainer to inspect the runner/model request logs or enable sanitized per-lens failure diagnostics in the trusted reviewer, then address the actual timeout/provider/response-validation failure and rerun the check.

PR #208, opened 2026-09-21
feat(third_party): link dependency-gateway as a git submodule

LJXLJXLJX:
## 1. Why the change?

The dependency gateway is the dependency cache backend of the image prebuild pipeline. Link it into agent-fleet as a submodule so the build-side infrastructure is discoverable from the monorepo while the service keeps its own repository, release cycle, and lint/test conventions.

## 2. Summary of the change

- Add `third_party/dependency-gateway` as a git submodule (github.com/sii-system/dependency-gateway)
- Register it in `AGENTS.md` / `STRUCT.md`

## 3. Other details

- No agent-fleet code paths change; prebuild keeps consuming the gateway via URL injection.
- The gateway is consumed by URL configuration only, so the submodule is optional for running tasks; CI checkouts (no submodules) are unaffected.


qiyuangong (approved):
LGTM

PR #211, opened 2026-09-21
fix(verifier): use portable static runtime bundle

LJXLJXLJX:
## 1. Why the change?

The SWE-rebench-V2 verifier bundle reused the Agent-side Python runtime, which is dynamically linked against host glibc. Task images may ship no Python, an old Python, or an incompatible glibc — in all these cases the verifier cannot start and the task fails for infrastructure reasons instead of model behavior.

## 2. Summary of the change

- Add `verifier_runtime/python_runtime.py`: a reusable, verifier-side runtime primitive that builds a fully static CPython 3.12.14 (x86_64-unknown-linux-musl, python-build-standalone, pinned SHA-256). The runtime wrapper clears `PYTHONPATH` / `LD_LIBRARY_PATH` / `LD_PRELOAD`, and the build validates source digest, ELF type (no PT_INTERP/PT_DYNAMIC), and required stdlib files. Non-x86_64 hosts fail explicitly; there is no fallback to the task image's Python.
- Rework `swe_rebench_v2_bundle_preparer` from a single module into a package that composes the static runtime into the verifier bundle and embeds a self-check (`bin/harbor-verifier-bundle-check` + `self_check.py`). `archive_ready()` now also verifies the self-check contents and interpreter symlinks.
- OpenSandbox command bootstrap (`yicloud_opensandbox.py`): run the bundle self-check after extraction, record its output (success and failure) in the bootstrap log, and exit 127 on failure before grading starts. Bootstrap log writes are now appended instead of overwritten.
- `dependencies.sh` / `environment_config.sh`: invoke the preparer as a package directory; update STRUCT.md and add READMEs documenting the verifier-runtime contract.

## 3. Other details

- Static source is only accepted from the local cache, `HARBOR_VERIFIER_PYTHON_SOURCE_ARCHIVE`, or the Artifact Cache Gateway; direct GitHub downloads through ambient proxies are not supported.
- Unit tests updated in `tests/test_swe_rebench_v2_bundle_preparer.py`; OpenSandbox smoke still to be validated separately.


qiyuangong (approved):
LGTM

qiyuangong:
@codex review

PR #212, opened 2026-09-21
chore(benchmark): drop redundant fixer report re-redaction

qiyuangong:
## Summary

Remove redundant redaction of the Fixer report inside `write_benchmark_summary.py`.

The benchmark summary writer runs `redact_sensitive_text()` over the Fixer
report text and stores it in `reports["fixer"]`, but the rendered `summary.md`
`## Fixer Results` section is rebuilt from the **original generated report**
(`_render_fixer_markdown(fixer_report_path, ...)` re-reads the file). The Fixer
already redacts at generation time in `harbor_fixer/report/markdown.py`, so
re-redacting the copy is dead work.

## Change

Fixer report text is now preserved verbatim in the summary input; redaction is
still applied to the Harbor and Analyzer reports, which are not redacted at
their source.

## Verification

- `test_benchmark_summary` (12), `test_run_summary` (12), `test_fleet_summary`
  (9), `test_harbor_summary_publishing` (5): all pass.
- `ruff check --config .github/ruff.toml` on the edited module: clean.
- One pre-existing failure in `test_harbor_fixer_report`
  (`test_generation_redacts_rerun_secrets`) is a macOS temp-path artifact
  (`/private/…` vs `…`) — reproduced on the unmodified upstream commit.


qiyuangong:
Closing this PR after a four-angle review (user / developer / architecture / devil's advocate).

**Conclusion: the premise was wrong.** `reports["fixer"]` is not dead code. It protects two active sinks — the persisted `benchmark-summary/summary-input.json` artifact and the Pi LLM prompt — and is used as the render gate. The Fixer's generation-time redaction is a strict superset of the writer's, so the double-pass was idempotent; removing it traded a near-free, effective defense-in-depth control to save one regex pass over a single report.

The one claim that survived all four reviews is a genuine, pre-existing gap — independent of this PR: `_render_fixer_markdown` pastes the fixer report's `Summary` / `Changes Applied` / `Remaining Issues` sections into `summary.md` with no consumer-side redaction, relying entirely on producer-side redaction. That is worth a follow-up (either a consumer-side redaction or an explicit producer-invariant assertion), but it has nothing to do with removing the payload redaction.

No behavior change was merged; the branch is closed unmerged. Notes for a follow-up:
- Add a regression test asserting no secret-shaped content appears in `summary-input.json`, the Pi prompt JSON, `summary-output.json`, or `summary.md` from any report (fixer included). No existing test guards this boundary.
- Consider rendering `summary.md`'s fixer section from the already-redacted payload text (`reports["fixer"]`) instead of re-reading the raw file, giving one source of truth.
- Consider a fail-closed invariant on the fixer path (assert `redact_sensitive_text(report) == report`) so a producer regression fails loudly instead of shipping verbatim text.

